PR #7, opened 2016-07-27
https support added

sk1talets:


src/common/request.js
```diff
@@ -54,6 +55,7 @@ export default class Request {
 
         request.end()
       } else {
+        const req = this._request
```
mefellows: I'm not super familiar with the browser request component of this, does it not also support `https`?

mefellows:
Thanks @sk1talets, it's looking good!

I'd suggest before merging that we also add an integration-style [test](https://github.com/pact-foundation/pact-js/blob/master/test/dsl/integration.spec.js) that spins up a Mock Server with SSL enabled and smoke test a basic scenario.

This will also likely require another configuration item for the Pact server to allow self-signed certificates (this would be the vast majority of cases i'd expect).

src/common/request.js
sk1talets: It does support https as far as I know. https://www.w3.org/TR/XMLHttpRequest/#introduction

sk1talets:
@mefellows, I did this already, if I got you right.. Take a look at `test/dsl/integration.spec.js`, I just run the whole suite for https:
```diff
-describe('Pact random mock port', () => {
+['http', 'https'].forEach((PROTOCOL) => {
+
+describe(`Pact random mock port, protocol: ${PROTOCOL}`, (protocol) => {
 
   const MOCK_PORT = Math.floor(Math.random() * 999) + 9000
-  const PROVIDER_URL = `http://localhost:${MOCK_PORT}`
+  const PROVIDER_URL = `${PROTOCOL}://localhost:${MOCK_PORT}`
   const mockServer = wrapper.createServer({
     port: MOCK_PORT,
     log: path.resolve(process.cwd(), 'logs', 'mockserver-integration.log'),
     dir: path.resolve(process.cwd(), 'pacts'),
+    ssl: PROTOCOL === 'https' ? true : false,
     spec: 2
   })
 
@@ -37,7 +40,11 @@ describe('Pact random mock port', () => {
 
   beforeEach((done) => {
     mockServer.start().then(() => {
-      provider = Pact({ consumer: `Consumer ${counter}`, provider: `Provider ${counter}`, port: MOCK_PORT })
+      provider = Pact({ consumer: `Consumer ${counter}`,
+                        provider: `Provider ${counter}`,
+                        port: MOCK_PORT,
+                        ssl: PROTOCOL === 'https' ? true : false
+                      })
```

mefellows:
@sk1talets Quite right LGTM - merging! We can look at adding https support for browsers in another PR. Much appreciated!

PR #9, opened 2016-08-08
fix(readme): fix example in readme

mefellows:
Fixes example in README so that it can be copy/pasted and run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,17 +70,14 @@ More questions about what's involved in Pact? [Read more about it](http://docs.p
 Check the `examples` folder for examples with Karma Jasmine / Mocha. The example below is taken from the [integration spec](https://github.com/pact-foundation/pact-js/blob/master/test/dsl/integration.spec.js).
 
 ```javascript
-import path from 'path'
-import { expect } from 'chai'
-import Promise from 'bluebird'
-import request from 'superagent'
-import wrapper from '@pact-foundation/pact-node'
-
-import { default as Pact } from 'pact'
-
-// great library to spin up the Pact Verifier Server
-// that will record interactions and eventually validate your pacts
-import wrapper from '@pact-foundation/pact-node'
+var path = require('path')
+var chai = require("chai")
+var expect = chai.expect
+var chaiAsPromised = require("chai-as-promised")
+var request = require ('superagent')
+var wrapper = require('@pact-foundation/pact-node')
+var Pact = require('pact')
+chai.use(chaiAsPromised);
 
 describe('Pact', () => {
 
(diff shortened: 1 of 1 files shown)


mefellows:
This doesn't require the use of babel or anything so I think it's probably a bit nicer - you can just run it with `mocha <filename>` and it should work. Thoughts?

tarciosaraiva:
Thanks @mefellows. Actually you might need babel-register depending on what you doing and the node version you're running. Node v4 implemented a good level of ES6 features and arrow functions are there so I believe it shouldn't be a problem.

As for the README, thanks for fixing that up - didn't realise there was a mistake there with two `it`. In any case please have a look at my comments on the file itself.

README.md
```diff
@@ -70,18 +70,18 @@ More questions about what's involved in Pact? [Read more about it](http://docs.p
...
-import Promise from 'bluebird'
-import request from 'superagent'
-import wrapper from '@pact-foundation/pact-node'
-
-import { default as Pact } from 'pact'
+var path = require('path')
+var chai = require("chai")
+var expect = chai.expect
+var chaiAsPromised = require("chai-as-promised")
+var Promise = require('bluebird')
```
tarciosaraiva: maybe we could remove this... avoid confusion?

README.md
```diff
@@ -70,18 +70,18 @@ More questions about what's involved in Pact? [Read more about it](http://docs.p
...
+var request = require ('superagent')
+var wrapper = require('@pact-foundation/pact-node')
+var Pact = require('pact')
+chai.use(chaiAsPromised);
 
 // great library to spin up the Pact Verifier Server
```
tarciosaraiva: get rid of these comments

README.md
```diff
@@ -129,15 +129,15 @@ describe('Pact', () => {
   })
 
   context('with a single request', () => {
-    it('successfully writes Pact file', (done) => {
+    describe('successfully writes Pact file', (done) => {
```
tarciosaraiva: `describe` should be used in sync mode. Remove the `done`

README.md
mefellows: Yeah I agree. It IS a great library though :)

README.md
mefellows: Done.

mefellows:
Thanks @tarciosaraiva. I've cleaned up as discussed. I can run the example with Node v4 (which is quite old now and the latest 6.x.x) with pure Mocha goodness.

tarciosaraiva:
Good stuff. Merging in.

PR #13, opened 2016-09-28
Better error handling in the example in case something goes wrong

jcarlos:
I was getting just timeout instead of a description of what was going wrong. This helped me debug.


tarciosaraiva:
nice thanks @jcarlos !

PR #19, opened 2017-01-18
Feat/e2e example

mefellows:
E2E working JS example, demonstrating the following Pact features:

* Consumer testing and pact file generation, including advanced features like:
    * [Flexible matching](https://docs.pact.io/documentation/javascript/flexible_matching.html)
    * [Provider states](https://docs.pact.io/documentation/provider_states.html)
* Sharing Pacts by publishing to and retrieving from a [Pact Broker](https://github.com/bethesque/pact_broker)
* Provider side verification

This comprises a complete E2E example that can be used as a basis for projects.


examples/e2e/.eslintrc.yml
```diff
@@ -0,0 +1,18 @@
+rules:
```
tarciosaraiva: so, by default we use [StandardJS](https://github.com/feross/standard)... this filed should not be needed.

examples/e2e/consumer.js
```diff
@@ -0,0 +1,82 @@
+const express = require('express');
+const request = require('superagent-bluebird-promise');
```
tarciosaraiva: you don't need that... Superagent 2 already has promises embedded into it. Unless you want to use something specific from Bluebird this is overkill

examples/e2e/consumerService.js
```diff
@@ -0,0 +1,8 @@
+let { server } = require('./consumer.js');
+
+server = server.listen(8080, () => {
```
tarciosaraiva: this association is kinda overkill, not sure why you would need that

mefellows: You can see that it's used below in outputting which host/ip it's bound to, but i'll simplify it.

examples/e2e/consumer.js
mefellows: Cool, will fix it. Don't need anything bluebird specific.

examples/e2e/.eslintrc.yml
mefellows: So just remove all rules and inherit from recommended? OK

mefellows:
Implemented feedback, thanks @tarciosaraiva - merging!

Issue #22, opened 2017-02-18
Provider Verification is broken


mefellows:
I noticed that the current `verify` function [expects](https://github.com/pact-foundation/pact-js/blob/master/src/pact.js#L80-L87) an http response. This is unusual, in most cases we will be calling service collaborators which in turn make the HTTP response calls on our behalf and return back an abstracted object (e.g. in a user service it might return a User object). This means we can only test trivial examples.

See https://github.com/pact-foundation/pact-js/blob/fix/verify/src/pact.js#L103-L116 for an example fix, part of a general API cleanup I'm undertaking with consultation with other community members.

Issue #21, opened 2017-02-17
Jest Example assertions never executed


thearegee:
#If you check the example: https://github.com/pact-foundation/pact-js/blob/master/examples/jest/__tests__/index.spec.js#L70

```Javascript 
it('successfully verifies', done => {
      return getMeDogs({ url, port })
        .then(provider.verify)
        .then(response => {
          expect(response.headers['content-type']).toEqual('application/json');
          expect(response.data).toEqual(EXPECTED_BODY);
          expect(response.status).toEqual(200);
        })
        .then(done, done)
    })
```
The done happens before the assertions are run, so these tests don't do anything, so it will always pass. If you refactor to look something like:

```Javascript
    it('successfully verifies', done => {
      return getMeDogs({ url, port })
        .then(provider.verify)
        .then(response => {
          expect(response.headers['content-type']).toEqual('application/json');
          expect(response.data).toEqual(EXPECTED_BODY);
          expect(response.status).toEqual(200);
          done();
        }, (error) => {
          console.log(error);
          done();
        })
    })
```

Now the assertions actually run however response is undefined so the tests fail. After some trial and error I found it was this line `.then(provider.verify)` made response return undefined. I'm not sure what that actually does but if you remove you get some passing tests.

```Javascript
    it('successfully verifies', done => {
      return getMeDogs({ url, port })
        .then(response => {
          expect(response.headers['content-type']).toEqual('application/json');
          expect(response.data).toEqual(EXPECTED_BODY);
          expect(response.status).toEqual(200);
          done();
        }, (error) => {
          console.log(error);
          done();
        })
    })
```

The problem with `.then(provider.verify)` is also true with the mocha test example. So my question is, what is `provider.verify` doing? Is it needed? If you let me know I can make a PR and fix these examples.

mefellows:
I've found this issue also (and actually properly raised it just now under #22). See https://github.com/pact-foundation/pact-js/blob/fix/verify/src/pact.js#L103-L116 for an example fix.

FWIW I'm working on an API cleanup as we speak, I hope to have this sorted in the next week or so.

PR #23, opened 2017-02-19
Updated API for 2.0.0 release

mefellows:
My take on an updated API that is simpler to use.

* Single dependency for Consumer and Provider verification, a common confusion for newcomers
* Fixes Pact Verification step (#21, #22)
* Retains Karma compatibility (via separate entry point `pact-karma.js`)
* API is similar, but definitely not backwards compatible so I suggest we bump to version `v2.0.0`

Almost all of the updated code are tests and documentation, but I see that as a good thing.


.istanbul.yml
```diff
@@ -8,8 +8,8 @@ check:
     functions: 80
     excludes: []
   each:
-    statements: 80
-    lines: 80
+    statements: 60
```
mefellows: I needed to do this because the Verifier class is basically a wrapper of `pact-node`'s Verifier and the integration tests didn't seem to cover it. Any unit test would stub it out such that it would be a useless test.

config/webpack.web.config.js
```diff
@@ -6,7 +6,7 @@ var DIST = path.resolve(__dirname, '../dist');
 var APP  = path.resolve(__dirname, '../src');
 
 module.exports = {
-  entry: path.resolve(APP, 'pact.js'),
+  entry: path.resolve(APP, 'pact-karma.js'),
```
mefellows: Similarly, as `pact-node` became a dependency, webpack pooped its pants and obviously couldn't deal with a few thousand Ruby files now on the path (and of course, this would never work in the browser). It seemed this was the cleanest option, albeit with some duplication.

examples/e2e/package.json
```diff
@@ -14,6 +14,7 @@
   "license": "MIT",
   "devDependencies": {
     "@pact-foundation/pact-node": "^4.6.0",
+    "cli-color": "^1.1.0",
```
mefellows: This is to make the console nice and red when there is a Pact failure (see https://github.com/pact-foundation/pact-js/pull/23/files#diff-ad2d27cf5c00ddf29bdafb5718a2a771R108). Might be overkill, thoughts?

thearegee:
Nice one, thanks for this

PR #35, opened 2017-03-28
Pass cors option to pact-node

jchen1:
This fixes a regression in pact 2.x where the pact-node CORS option was no longer exposed


gcarling:
+1

mefellows:
Thanks for this, just releasing `v2.3.1` now (build on its way).

PR #36, opened 2017-04-06
core(docs): remove anti-pattern from example in readme

BenSayers:
The example Pact test in the README contains a test anti-pattern. The call to `provider.verify` is in a seperate test to the code that actually makes the http request that is verified. This leads to brittle tests as a change in the order of test execution will cause the test to fail. For example, jasmine has a flag that will cause it to execute tests in a random order.


mefellows:
Thanks Ben - sorry for my slowness to respond, have been afk.

PR #48, opened 2017-05-09
feat(mock service): add pactfile_write_mode option handling

Narazaka:
I want to use the [pactfile_write_mode feature](https://github.com/realestate-com-au/pact/blob/master/documentation/configuration.md#pactfile_write_mode) for separated test files.

### DIFF
--- a/src/dsl/mockService.d.ts
+++ b/src/dsl/mockService.d.ts
@@ -1,7 +1,16 @@
 import {Interaction} from './interaction';
 
+export type PactfileWriteMode = "overwrite" | "update" | "smart" | "none";
+
 export class MockService {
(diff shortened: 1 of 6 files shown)


mefellows:
This is a great idea!

Have you tested that this gets through to the underlying Ruby process, though (i.e. have you tested this works).

Likely, this would also need to go into http://github.com/pact-foundation/pact-node also, if not into the CLI of the Ruby process.

Narazaka:
I had tested it with "overwrite" and "update" option values and checked their working fine. The "pactfile_write_mode" option is handled by not cli options but pact mock server's POST /pact API.
The "smart" option seems to depend on rspec so I ignored that.

mefellows:
Awesome - I'll bring it in, update the docs etc.

This will be useful for a number of use cases!

PR #62, opened 2017-06-27
Updated documentation around like and eachLike

bethesque:
Putting a `like` around each value is tedious. The documentation needs to show that the like can be wrapped around the entire object.


mefellows:
Thanks, I actually wasn't aware of this until recently. We should also update the E2E tests.

PR #72, opened 2017-07-28
docs(README): updating example

lirantal:
* Removing unused `request` library being required
* The port constant was not defined
* Cosmetic change from `let` to `const`


mefellows:
Thanks!

PR #74, opened 2017-07-29
docs(readme): improved example for asserting expectations on contract

### LINKED ISSUE #65: Question on provider.verify() in consumer test?
In examples/e2e/test/consumer.spec.js, I find the below test. 
```
describe('when interacting with Animal Service', () => {
    it('should validate the interactions and create a contract', () => {
      return provider.verify()
    })
  })
```

**Question: I don't understand what "provider.verify()" does? Is this mandatory to be included in a test? or I can remove it. After I remove it, I can still see log and pact file are generated and test is run as well.** 

I have a test like this:

```
describe('CRUD operations:', () => {
    describe('POST',() => {
      before((done) => {
        provider.addInteraction().then(() => { done(); });
      });

      it('POST', (done) => {
            done();
          })
          .catch(done);
      });

      it('successfully verifies', () => provider.verify());
    });

    describe('PUT',() => {
      before((done) => {
        provider.addInteraction().then(() => { done(); });
      });

      it('PUT', (done) => {
            done();
          })
          .catch(done);
      });

      it('successfully verifies', () => provider.verify());
    });

  });
```

lirantal:
Fix #65 by asserting that the promise for `provider.verify()` hasn't thrown an error, thus providing a clearer example of usage for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,8 +172,9 @@ describe('Pact', () => {
       	    expect(projects).to.be.a('array')
             expect(projects).to.have.deep.property('projects[0].id', 1)
 
-            // (5) validate the interactions occurred, this will throw an error if it fails telling you what went wrong
-      	    return provider.verify()
+            // (5) validate the interactions you've registered and expected occurred 
+            // this will throw an error if it fails telling you what went wrong
+      	    expect(provider.verify()).to.not.throw()
           })
       })
 


lirantal:
@mefellows what do you think?
I'm wondering if it's also better to update the API table entry to verify to make it clearer that this is a promise that should be asserted, or do you think the example is enough?

mefellows:
Why not both?

ef4ecda158ef1e79079546811ace2017bd04708149857b1cc6ff5d3712c11d5f

lirantal:
@mefellows don't worry, ~~winter~~ more is coming!

PR #82, opened 2017-08-08
feat(pact-server): pact server supports config from env

lirantal:
Supports providing Pact Server configuration through environment variables

### DIFF
--- a/src/common/config.js
+++ b/src/common/config.js
@@ -11,8 +11,8 @@ const resolve = require('path').resolve
 const PACT_CONFIG_FILE = resolve(process.cwd(), 'config', 'pact.config.js')
 const PACT_CONFIG = {
   mockService: {
-    host: '127.0.0.1',
-    port: 1234
+    host: process.env.PACT_SERVER_ADDR || '127.0.0.1',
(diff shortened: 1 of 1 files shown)


mefellows:
It's a good idea and looks OK to me, I'll have a broader look later today - perhaps we can make this extend into other parts of the configuration.

lirantal:
Cool.

bethesque:
This assumes you're only using one mock server.
It would be more accurate to call it PACT_MOCK_SERVICE_HOST and PACT_MOCK_SERVICE_PORT because the PACT_SERVER is ambiguous. Is it the mock server, the server you're verifying, or the server you're retrieving the pacts from?

lirantal:
Agree on the naming convention.

With regards to the mock server, I think the env params can be used as defaults for when the mock server provider is bootstrapped such as:
```
pact({
  consumer: '',
  provider: '',
  host: '',
  port: ''
})
```

mefellows:
Hi @lirantal, I'm going to close this as it is embarrassingly overdue (from our end). I'll raise a separate issue to capture the intention of this where we can discuss the value in doing an ENV based config.

PR #87, opened 2017-08-19
docs(verify): better describe the verify method

lirantal:
In response to our discussion on gitter with @mefellows and @bethesque about how the verify method works.


mefellows:
thanks!

bethesque:
I'd prefer not to mention the query parameters! That's probably a bug. Also, this wording very unspecific. If you specify parameters, and they don't match, the request will not match, and that is not clear from this wording.

bethesque:
I've found some documentation on request matching that I'd forgotten about. I've included a link to it in that section.

lirantal:
@bethesque I'm more than happy to correct. What's the link to the documentation you found? and also let's wait to see if this is a bug or what :)

bethesque:
I've updated that section, you'll see the link there now.

lirantal:
super! looks good.

PR #95, opened 2017-09-13
The DSL matcher eachLike should also accept empty arrays

ptornhult:
I've changed the DSL matcher eachLike to accept empty arrays. This was allowed in Java Pacts so I'm assuming it's also allowed in JS Pacts. I've added Jest tests for each matcher so others can learn by example as well. I could probably add AVA and Mocka tests as well but I'd like to check if there is any response on my PR before putting in the work ;)


bethesque:
@uglyog and I disagree about allowing users to shoot themselves in the foot :P

I don't think this should be allowed, because all your pact tests could pass, but the service only provided an empty array, and your item handling code never got executed. If you need to test both code paths, I suggest you make one example for an empty array, and one example with at least one item.

ptornhult:
I'm not sure what you are saying. Are you saying that empty arrays shouldn't be allowed, or that I should add another test so that the example would cover both empty and filled array?

bethesque:
I think you should be able to specify a test with an empty array, and a separate test with *one or more* items However, a single test with zero or more items gives you no confidence about what the actual items are. You don't know if it passed because there was an empty array returned, or because items matching your example were returned.

ptornhult:
I've added another interaction to test non-empty arrays as well, but this won't enforce others to do the same. Hopefully people can follow the examples, but I'm not sure what else can be done to prevent people from skipping this test and making mistakes in the future?

mefellows:
I think what Beth is saying, is that in your Pact tests for your project you should create two tests:

* one for the state where (to continue your analogy) there are no cats and you get back `[]`
* another for the state where there are cats and you get back an array of cats `[{...}]`

The schrodinger analogy is pretty good here - if you expected an array that _could_ contain a cat, but also could contain an array of length zero, then at provider verification time an empty array will pass the test. At runtime though, when you opened the proverbial box, you _could_ get back an array of anything which could break your consumer.

bethesque:
Exactly. To follow the analogy, you could open the box in production and get an array of dogs.

ptornhult:
I've already added another test to cover the non-empty array scenario in the Jest examples. All I'm saying is that it won't force anyone to do the same in their tests.

What is the way forward here, what do you want me to do?

bethesque:
Hi Peter. Thank you for your contribution, however, I am not going to accept this PR. Our comments above are not about your tests for the requested feature, which are fine. It's about the functionality of the feature itself. It is not a feature that I am going to allow in the pact library, for the reasons that Matt and I have explained. It would allow users to write tests that hid bugs that might not show until production. 

One thing to keep in mind about pact is that you are writing examples, not schemas. A schema describes all possible states of a document, and hence, it makes sense to be able to specify that an array may have 0 => many items. Schrodinger's box is not yet open. With pact, you are specifying concrete examples. The same concrete example cannot have both an empty array and a populated array. Running the example collapses the waveform, and at that stage, you have to choose between whether you expect an empty array, or a populated array.

Please let me know if I've made that clearer or more confusing!

ptornhult:
Thank you, I understand. 

But then I'm curious if this should still be allowed in the JVM version ([See eachLike, line 646, **matchMin(0)**](https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonBody.java)). 

Maybe that should be changed to only accept min 1 element as well in that case. I'll file an issue on that and see if I get any response there. It doesn't give me confidence that the different language implementations of DSL behave differently.

Thank you for taking the time to respond!

mefellows:
Yeah, I think it's worth looking at making this consistent (whichever side of the line we decide to go). Compatibility is important.

PR #101, opened 2017-09-29
fix(src): fix typo in error message

ptornhult:
fix a typo in eachLike DSL matcher were comparison to < 1 would cause an error but error message
said > 1 which is incorrect


mefellows:
Thanks!

PR #114, opened 2017-10-15
fix(verification): pass validation error message on 

### LINKED ISSUE #113: verification error message is swallowed
When a consumer verification fails, the error message is logged, but not passed upwards to the caller. This means that you always have to look into the logs to find out what went wrong. 

For example, using `karma-pact`, when the tests are run in the browser, you get the output "verification failed" in the browser, but nothing else. Especially when you're starting with Pact, this makes for long bug hunts. It would be nice if the error message directly states what part of the verification failed.

thombergs:
Verification error message is now passed on instead of being swallowed.

I'm not sure if it has any side effects, but at least for my use case (through karma-pact) this change made the experience of writing consumer tests much more pleasant, since I can now see what went wrong in the karma results in the browser.

fixes #113

### DIFF
--- a/src/pact-web.js
+++ b/src/pact-web.js
@@ -75,12 +75,7 @@ module.exports = (opts) => {
       return mockService.verify()
         .then(() => mockService.removeInteractions())
         .catch(e => {
-          // Properly format the error
-          console.error('')
-          console.error('Pact verification failed!')
-          console.error(e)
-
-          throw new Error('Pact verification failed - expected interactions did not match actual.')
+          throw new Error(e)
         })
     },
     /**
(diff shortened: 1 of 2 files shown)


mefellows:
It makes sense. Could you please elaborate as to why the stack trace should be removed? It's typical of a test framework to output the results to a terminal, but I suspect perhaps this is less helpful in the Karma-like use cases?

thombergs:
Took a couple commits to fix the test, because I couldn't get them to run on Windows (might make a separate PR for that), so I force pushed a clean history. The CI build still fails however, but because of some other reason.

I removed the error log since I assumed whatever framework is calling the verify() function will output the error as it sees fit and I don't see why it should be logged twice. 

I'm not very deep into the JS ecosystem, however, so I don't really know the best practice here :).

mefellows:
Thanks. Not sure what is going on with the Node 4.x.x build but it's related to the karma examples I updated in master last night, not your change which is good to bring in.

Let's see how this goes!

PR #128, opened 2017-11-27
build(travis): make sure we test all examples code

lirantal:
I noticed we're not checking the actual examples code to make sure they pass and actually work so I updated the build script which is used for the e2e examples.


bethesque:
Awesome!

mefellows:
I remember now, we need Node 4 builds to pass 👍

lirantal:
@mefellows what's failing the build now?

mefellows:
Looks like a recent version of `jscpd` is not Node 4 compatible. Will fix.

Issue #10, opened 2016-08-31
mockServer does not `start()`


dolfo-pivotal:
Hi, 

I am trying to start doing CDD on my JS application. I am trying to create the most simple consumer pact, but when I call 

```
mockServer.start().then(() => {
    provider = Pact({consumer: 'client', provider: 'service', port: 1234});
   done();
});
```

It never gets inside the `then`, so I am assuming the promise never gets resolved. I am using Jest that uses Jasmine behind the scene. I wonder if you have any clue of how could I debug this network issue ( if it is that ).

this is my test file 

```
import Pact from 'pact';
import path from "path";
import chaiAsPromised from "chai-as-promised";
import pactWrapper from "@pact-foundation/pact-node";
import chai, {expect} from "chai";
import request  from 'superagent';

chai.use(chaiAsPromised);

describe("see all dog", () => {
    var provider, mockServer;

    beforeAll(() =>{
        mockServer = pactWrapper.createServer({
            port: 1234,
            host: "localhost",
            log: path.resolve(process.cwd(), 'logs', 'mockserver-integration.log'),
            dir: path.resolve(process.cwd(), 'pacts'),
            consumer: "client",
            provider: "service",
            spec: 2
        });
    });

    afterAll(() => {
        pactWrapper.removeAllServers();
    });

    beforeEach((done) => {
        mockServer.start().then(() => {
            provider = Pact({consumer: 'client', provider: 'service', port: 1234});
            done();
        });
    });

    afterEach((done) => {
        mockServer.delete().then(() => {
            done();
        });
    });

    describe("test", () => {
        describe("test", () => {
            beforeEach((done) => {
                provider.addInteraction({
                    state: 'I want to see my dogs',
                    uponReceiving: 'a request to get dogs',
                    withRequest: {
                        method: 'GET',
                        path: '/dogs',
                        headers: {'Accept': 'application/json'}
                    },
                    willRespondWith: {
                        status: 200,
                        headers: {'Content-Type': 'application/json'},
                        body: {}
                    }
                }).then(() => done());
            });

            afterEach((done) => {
                provider.finalize().then(() => done())
            });

            it('successfully verifies', (done) => {
                const verificationPromise = request
                    .get('http://localhost:1234/dogs')
                    .set({ 'Accept': 'application/json' })
                    .then(provider.verify);

                expect(verificationPromise).to.eventually.eql(JSON.stringify({})).notify(done)
            })
        });
    });
});
```

and when I run test I get 

```
[...]

PR #134, opened 2017-12-08
feat(api) allow overriding what kind of request is constructed

mbark:
First of all I just wanted to say that when going through the code it is very legible and well-written!

## Background
We use `jest` to test our `angular` application and want to introduce pact-testing for the frontend-backend-communication. However, we run into the following problem:
- In order to run `angular` we need to have our environment set to `jsdom` due to `zone.js` patching XMLHttpRequest etc;
- In order to run `pact` we need to have our environment set to `node` (see #10 for further discussion and background).

I modified the code so that we always used `node`-style requests (ignoring environment checks) which got it working for us. With that in mind I thought it might be possible to add an ability to override the default behaviour, allowing us in scenarios such as these to run in the `jsdom` environment but tell `pact` to use `node`-style requests.

## System solution
This feature introduces an optional map of values to give when setting up the pact server. For now these options only contain one possible key: `environment`. The environment option can be used to force pact to use either a node-style request or a web-style request.

As part of this I also added `xhr-mock` and the ability to mock the `window`, which allowed running the test for `Request` preferring the `XMLHttpRequest` object.

If you think this is a good solution to go ahead with I can improve the tests a bit and document this new functionality in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ The `Pact` class provides the following high-level APIs, they are listed in the 
 | `spec` | no | number | Pact specification version (defaults to 2) |
 | `cors` | no | boolean | Allow CORS OPTION requests to be accepted, defaults to false |
 | `pactfileWriteMode` | no | string | Control how the Pact files are written. Choices: 'overwrite' 'update' or 'none'. Defaults to 'overwrite'|
+| `environment` | no | string | Control which environment pact will consider itself being ran in. Choices: 'web' or 'node'. Defaults to identifying environment by checking for the existence of `window` or `XMLHttpRequest` |
 
 #### Example
 The first step is to create a test for your API Consumer. The example below uses [Mocha](https://mochajs.org), and demonstrates the basic approach:
--- a/package.json
+++ b/package.json
@@ -140,6 +140,7 @@
     "ts-node": "^3.3.0",
     "tslint": "^5.8.0",
     "typescript": "^2.4.2",
-    "webpack": "^3.5.5"
+    "webpack": "^3.5.5",
+    "xhr-mock": "2.0.0-preview.10"
   }
 }
--- a/src/common/environment.spec.ts
+++ b/src/common/environment.spec.ts
@@ -1,0 +1,32 @@
+/* tslint:disable:no-unused-expression no-empty */
+import * as chai from "chai";
+import * as sinon from "sinon";
+import mock from "xhr-mock";
+const expect = chai.expect;
+import { environment } from "./environment";
+
+describe("Environment", () => {
+  context("#isBrowserEnvironment", () => {
+    const g: any = global;
+
+    beforeEach(() => {
+      g.window = {};
+      mock.setup();
+    });
+
+    afterEach(() => {
+      mock.teardown();
+      g.window = undefined;
+    });
+
+    it("should consider it to be a browser environment", () => {
+      expect(environment.isBrowser()).to.be.true;
+    });
+  });
+
+  context("#isNodeEnvironment", () => {
+    it("should consider it to be a node environment", () => {
+      expect(environment.isNode()).to.be.true;
+    });
+  });
+});
--- a/src/common/environment.ts
+++ b/src/common/environment.ts
@@ -1,0 +1,12 @@
+const environment = {
+  isBrowser: (): boolean => {
+    return (
+      typeof window !== "undefined" && typeof XMLHttpRequest === "function"
+    );
+  },
+  isNode: (): boolean => {
+    return typeof module !== "undefined" && !!module.exports;
+  },
+};
+
+export { environment };
--- a/src/common/request.spec.ts
+++ b/src/common/request.spec.ts
@@ -3,7 +3,10 @@ import * as chai from "chai";
 import { loadavg } from "os";
 import * as sinon from "sinon";
 import { parse, Url, URL } from "url";
+import mock from "xhr-mock";
 const expect = chai.expect;
+import { Environment } from "../dsl/options";
+import { environment } from "./environment";
 import { Request } from "./request";
 
 describe("Request", () => {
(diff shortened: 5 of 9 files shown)


mefellows:
Hi @mbark, i'm so sorry for the delay in responding to this. Not sure how it didn't get on my radar (perhaps when I was on parental leave I saw it but didn't respond - crazy time), but in any case let's get into it and thanks for the kind words!

OK so if I understand correctly, the root cause of your problem is this line: `if (typeof XMLHttpRequest === "function" || typeof window !== "undefined") {`.

Because you are patching with libraries, this statement will never reflect the "true" environment and the pact library may not behave as expected, so we want to maintain complete control if possible, or default to the current behaviour if the user doesn't specify a preference?

In theory, I support this change.

In terms of implementation, here are my additional thoughts:

* The assumption is that this functionality is not covered by either the `pact` (designed for node environments) or`pact-web` module (designed for non-node/browser environments)
* Where possible, maintain backwards compatibility in types - I'd prefer not to bump a major version for this feature, so any new parameters must be optional in any public signatures/constructors, or optional in existing types (e.g. `PactOptions`)
* Rather than have a separate parameter to the `Pact` constructor, add `environment` as an optional property in the `PactOptions`type (instead of `AdditionalOptions`).
* The use of `xhr-mock` is 👍

mbark:
I rebased it against master and will go through your comments later today (or at the very latest this week)!

mbark:
I added `environment` to `PactOptions` and made it optional. I also added it to the table describing the `PactOptions` parameters. Not entirely sure about the phrasing there to make it most clear how it works though.

Maybe you could take a look and see what you think?

It also seems like `travis` broke when installing node 8, which is not really related to what I have done at all. The other node versions work and so does 8, that was the version I tested against locally.

mefellows:
Thanks @mbark, I'll look into this over the weekend but on first glance it looks promising!

mefellows:
Added @mboudreau for his thoughts on this change also.

mbark:
@mefellows is there any progress on getting this merged? Or anything more I can do to help getting it merged? :)

mboudreau:
Sorry, I needed to look this over. My bad.

mboudreau:
@mbark Thanks for your PR.  I had a bit of a look and it looks mostly good except for a few opinionated things that I have, but before I want to get into that, I'm trying to understand your scenario better so that we can make the best call possible for the library and our users.

From what I understand (correct me where I'm wrong), but you want to run pact with your jest tests, but because of your environment setup, Request is confusing which actual environment it's in (web vs node) and causing some issues when the requests are actually built.  You've found a workaround for now, but it isn't elegant.

I don't know jest that well, so I'm guessing it's a node based testing solution for React, but you want it to run for angular, hence the need to "fake" the environment with jsdom, causing the issue in the first place.

I want to try out a few solutions to see what works best in development.  Do you have a very small sample project you can setup using your same scenario, that way I can test out to make sure it works for everyone.  

Thanks.  I promise that we'll find something that works for everyone :)

mbark:
@mboudreau Regarding opinions about the code structure/names etc I'm all for you just making the changes you want to -- I'm interested in getting the feature merged, not the exact implementation of it. We can take that later though!

When it comes to your understanding of the problem I'd say that it is correct and I'll try to take some time later today to re-create our problem.

mbark:
@mboudreau I have reproduction at: https://github.com/mbark/pact-jest

Run `yarn jest --projects jest/test.config.js` to run the tests (both pact and the dummy `app.spec` one). This should give you the error:
```

    console.error node_modules/jsdom/lib/jsdom/virtual-console.js:29
      Error: Error: connect ECONNREFUSED 127.0.0.1:1234
          at Object.dispatchError (/Users/martin.barksten/repos/pact-jest/node_modules/jsdom/lib/jsdom/living/xhr-utils.js:65:19)
          at EventEmitter.client.on.err (/Users/martin.barksten/repos/pact-jest/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:664:20)
          at EventEmitter.emit (events.js:164:20)
          at Request.preflightClient.on.err (/Users/martin.barksten/repos/pact-jest/node_modules/jsdom/lib/jsdom/living/xhr-utils.js:409:47)
          at Request.emit (events.js:159:13)
          at Request.onRequestError (/Users/martin.barksten/repos/pact-jest/node_modules/request/request.js:878:8)
          at ClientRequest.emit (events.js:159:13)
          at Socket.socketErrorListener (_http_client.js:389:9)
          at Socket.emit (events.js:159:13)
          at emitErrorNT (internal/streams/destroy.js:64:8) undefined
```

If you then edit `node_modules/@pact-foundation/pact/common/request.js` and change:
- `typeof XMLHttpRequest === 'function' || typeof window !== 'undefined'`  to `false`;
- `typeof window === 'undefined'` to `true` (both in the constructor and the `send` method).

Both tests should pass. You can also try setting `@jest-environment node` in the `pact.spec.ts` file to see the error when running in `node` rather than `jsdom` mode.

Hopefully that works as a simple reproduction of the problem :)

mboudreau:
Amazing work, thank you. This is very helpful. I'll try to have everything
sorted out by tomorrow. Cheers.

mboudreau:
@mbark Hey man, I'm looking into using an external library that would handle both node/browser requests called [Popsicle](https://github.com/blakeembrey/popsicle).  I just tested it using your example project (gj on that btw) which seems to fix the issue.  I'm going to add it to the codebase and create a new PR for it and close this one once it's gone through.  Cheers and thanks for the help :)

mboudreau:
@mbark I've create a new PR with my proposed changes, let me know what you think https://github.com/pact-foundation/pact-js/pull/147

mefellows:
Thanks for this @mbark, will be closing this in favour of #147

PR #141, opened 2018-01-05
Added jest multiple pact test files example

elliottmurray:
I could probably add some docs to why this is so? And some other tips (like running pact tests as a separate task so that you can run normal unit tests concurrently)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,8 @@ pacts/
 .idea
 *.iml
 .vscode
+*.swp
+
+# Local config files
+.nvmrc
+.envrc
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ pacts/
 # Local config files
 .nvmrc
 .envrc
+
--- a/examples/jest/README.md
+++ b/examples/jest/README.md
@@ -3,3 +3,18 @@
 1. In the pact-js project root, change to the `examples/jest` directory
 1. Run: `npm i`
 1. Run the tests: `npm t`
+
+## Comments about Jest
+You will need to run jest 'in band' as it will cause state issues otherwise. If you are running a large unit test suite you'll probably want to run that separately as a result to take advantage of the concurrency of jest (it is quite a slow down). To achieve this you can get your pact tests to have a suffix of '.pact.js' and add the following to your pact task in npm:
+```
+--testRegex \"/*(.test.pact.js)\""
+```
+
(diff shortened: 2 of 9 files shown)


lirantal:
this is great @elliottmurray , thanks for jumping in on it!
some comments:
* adding an interaction, testing it, and verifying it should all happen within the test case itself. meaning you need to remove addInteraction() from the beforeAll() clause

elliottmurray:
Hey,

I actually based this on the current jest example. But now having changed it to add the same interaction in each spec it fails saying missed interaction. I think because it now is looking for the same interaction twice now? I've tried a series of other things but the lifecycle of jest means I'm struggling to get multiple test files to work if I add each interaction more than once in total (hence why the beforeAll in the specific describe block)

Suggestions? 

I'll push some README instructions too 

screen shot 2018-01-06 at 11 09 44 am

mefellows:
FYI I've fixed the TS issue in master, if you rebase your changes on that the build should pass.

mefellows:
I'll pull your examples down now @elliottmurray, but one thing I'll say is that I'm not tied to the contrived example (I can't recall the history of this example) - if you wanted to do something else go for it.

elliottmurray:
Cool - I'll have another play next week. Didn't want to have those add interactions in the beforeAll but can't get the provider to play nicely.

mefellows:
@elliottmurray I've taken a look at your example, looks like if you did something like this everything will be hunky dory. Note how the verification happens as part of the individual test (guaranteeing order etc.):

```js
    // add expectations
    it('returns a sucessful body', () => {
      return getMeDogs({
        url,
        port
      })
        .then(response => {
          expect(response.headers['content-type']).toEqual('application/json')
          expect(response.data).toEqual(EXPECTED_BODY)
          expect(response.status).toEqual(200)
        })
        .then(() => provider.verify())
    })
```

Alternatively, an `afterEach` could probably do the trick, but I'm not as deeply aware of the Jest semantics (e.g. does it also run after other types of functions, including setup/teardown ones) i.e.

```
afterEach(() => {
  console.log("VERIFYING!")
  return provider.verify()
});
```

That works nicely in your `pactTestWrapper.js` file too.

elliottmurray:
I've made that change for the verify as part of the promise chain. Also changed to a beforeEach and added there which I think is fine personally. But can move it to the specific test if required.

mefellows:
Thanks Elliot, when the tests pass I'll merge this in!

mefellows:
ok, build needed to be re-run but it's green - coming in!

PR #147, opened 2018-01-31
using Popsicle for cross node/browser requests

mboudreau:
seems to work fine, need to refactor tests, but frankly, not sure if we actually need tests for the requests since they are already tested from the library.  Might need more integration like tests with pact though.


mboudreau:
@mefellows your linting is ludicrous :P

tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "target": "es5",
     "typeRoots": [
       "node_modules/@types",
-      "src/types/**/*"
+      "src/types/**/*",
+      "node_modules/popsicle/**/*"
```
mbark: Do these work for you? I don't get popsicle typings working neither with nor without these `typeRoots`.

mefellows:
Nice work guys, I'll try and take a look at this today some time (on train etc.). On first glance, it looks OK.

Yes, we could probably look at linting (the parameter ordering is the most painful and probably unnecessary).

tsconfig.json
mboudreau: Sorry, forgot I had added that.  Removed.

src/common/request.ts
```diff
@@ -1,91 +1,38 @@
 "use strict";
 
-import { parse, Url } from "url";
-import { logger } from "./logger";
+import * as Popsicle from "popsicle/dist/common";
```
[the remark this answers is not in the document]
mboudreau: It's funny, because it compiles fine on my computer, but travis keeps failing because of some of the dependencies in popsicle.  Need to investigate further.

mboudreau:
@mbark You'll have to excuse me on this, I've been putting out fires all week. This will get the attention it needs.  I'll try to get to it this weekend.

mbark:
@mboudreau No problem! There is no hurry as long as things are moving forward :) If you need any help getting it merged I could check it out over the weekend as well.

src/common/request.ts
```diff
@@ -1,8 +1,8 @@
 "use strict";
 
 import * as Superagent from "superagent";
-import {logger} from "./logger";
 import {Response} from "superagent";
+import {logger} from "./logger";
```
mefellows: If you use a 'real' IDE it will do this for you 😆

mboudreau:
gotcha.  I can rebase it and force push for merging purposes.  Also, my only concern is that superagent (I moved away from popsicle, it had some issues with typescript which was annoying me) might not be able to handle https request for insecure certificates, which is a node thing, and might happen to our users.  I need to create some tests around that to make sure it still works, and if not, if there's a way to get around it.

mefellows:
It's a very good point. I'm not sure how we dealt with it using `XMLHttpRequest`, if at all.

Because it would have been running in a browser, presumably it would require user input/overriding in session, something that I can't see being allowed in code, because that would mean any in-browser security attacks could privilege escalate.

My guess is that some flag (e.g. `--ignore-ssl-errors`) would need to be provided to the running browser via a configuration (e.g. karma config).

mboudreau:
@mefellows that's correct, or that the user 'accepts' the cert in the browser beforehand, but that's a fairly well known "issue" with a very clear way of fixing it in the browser, but this is not possible with node within either setting an environment variable or specifying it in the http configuration before calling it.

In the past, we've specifically ignored all cert issues by setting a flag, however, the more I think about it, the more I believe we shouldn't be messing with security in this way.  Maybe as a better approach is to allow the developer to specify either the cert/key to allow in pact-js, so that it can be passed onto superagent, or have the developer set the node environment variable to ignore SSL certificate issues.

Thoughts?

mefellows:
I'm with you - I think pact-node sets that flag which is not at all obvious. We should either have a flag that says "allow invalid certs" or provide the certificate infrastructure to make it work.

mboudreau:
yeah, I think I'm agreeing with you there.  So, how do we move forward with
this one then?

mefellows:
My preference would be to pop this behaviour behind a flag that disables ssl validation, however given your comments and a cursory glance at superagent I'm not sure the best approach.

It seems possible to set the node env var (see this related [pact-node issue](https://github.com/pact-foundation/pact-node/issues/61) I raised a while back), so perhaps that is the simplest path forward? If this flag is set, we disable SSL security for the session.

This would presumably also open up the possibility to modify the SSL runtime, specifying CA bundles, certs, keys etc.

Seems like a bit of work though - perhaps a PR into superagent is really the ultimate solution.

Or, deferring back to Popsicle as it has what we need (minus the Type niceties?)

mboudreau:
I guess it depends what we want to allow pact-js to do and externalize.  A simple workaround would be to set the node environment variable right before the call using superagent, then removing it after the call comes back.  There might also be a way to modify the https dependency itself, not sure.

I'm okay with just disabling SSL errors altogether since this isn't meant to be used to validate certs.

mefellows:
Yep, the whole testing https thing in pact tests is really tangential to this whole process - you basically learn nothing from it. I think your suggestion is reasonable, provided it is kept behind a flag and user configurable (so they know "I'm about to do dangerous and also not very useful stuff").

mboudreau:
@mefellows Alright, I think I got popsicle to work.  Superagent's creator was extremely strongly opposed at having an SSL ignore option.  I still don't have any tests, but I've setup a few empty ones with the description of what I want it to do.  I'm going to use nock to create mock APIs so we can test the Request class in isolation.

mefellows:
Awesome work, thanks @mboudreau. Yeah, I read his responses on the GH issues. I can understand his point, but for our use case I think those risks are acceptable.

tslint.json
```diff
@@ -6,7 +6,8 @@
   "jsRules": {},
   "rules": {
     "interface-name": [true, "never-prefix"],
-    "no-var-requires": false
+    "no-var-requires": false,
+    "ordered-imports": false
```
mbark: A good way to do this rather than disabling it is to add [import-sort](https://github.com/renke/import-sort) as a pre-commit hook (via `husky` + `lint-staged`) to automatically sort all imports.

mboudreau:
@mefellows @mbark I think I'm finally done this.  You'll see that this change includes a few extra things as I got a bit carried away trying to fix things as I was seeing them.  It passes all tests and I think will make the code moving forward quite a bit cleaner.  Let me know what you think.

mefellows:
Thanks @mboudreau! I'll take a look tonight (if possible).

src/dsl/interaction.spec.ts
```diff
@@ -1,12 +1,16 @@
 import * as chai from "chai";
-const expect = require("chai").expect;
-import { Interaction } from "./interaction";
+import * as chaiAsPromised from "chai-as-promised";
+import {HTTPMethod} from "../common/request";
+import {Interaction, RequestOptions} from "./interaction";
```
mefellows: I've just pushed up an updated linting config that should ensure consistency of the padding around braces (amongst others). I'm guessing your editor doesn't like the padding.

src/dsl/interaction.ts
```diff
@@ -80,23 +79,15 @@ export class Interaction {
...
     }
 
-    this.state.request = omitBy({
-      body: requestOpts.body,
-      headers: requestOpts.headers,
-      method: requestOpts.method.toUpperCase(),
-      path: requestOpts.path,
-      query: requestOpts.query,
-    }, isNil) as RequestOptions;
+    this.state.request = omitBy(requestOpts, isNil) as RequestOptions;
```
mefellows: DUH!

src/pact.integration.spec.ts
```diff
@@ -267,49 +257,35 @@ describe("Integration", () => {
...
-        });
-
-        it("fails verification", (done) => {
-          const promiseResults: Array<Promise<any>> = [];
+            body: EXPECTED_BODY,
+          },
+        }).then(() => console.log("Adding interaction worked"), () => console.warn("Adding interaction failed.")));
 
+        it("fails verification", () => {
+          // TODO: fix this test, not sure what it's trying to do
```
mefellows: I think you've captured the intention. I assume it was about making two requests but only supplying one. I think we can remove the TODO.

src/pact.ts
```diff
@@ -146,10 +150,19 @@ export class Pact {
     this.finalized = true;
 
     return this.mockService.writePact()
-      .then(() => this.server.delete())
-      .catch((err: Error) => {
-        return Promise.all([this.server.delete(), Promise.reject(err)]);
-      });
+      .then(() => logger.info("Pact File Written"), (e) => {
+        console.log("Pact File Write Failure");
```
mefellows: Any reason for the `console.log` independent of the log?

src/pact.ts
```diff
@@ -146,10 +150,19 @@ export class Pact {
...
+        return Promise.reject(e);
+      })
+      .then(() => new Promise<void>((resolve, reject) => this.server.delete().then(() => resolve(), (e) => reject(e))))
```
mefellows: This is a bit hard to read/reason - worth extracting?

src/pact.ts
mefellows: Are they just debug statements during dev that needed to go?

mefellows:
FWIW Node 8 locally also hangs for me @mboudreau. I'll try to resolve it this morning on the train.

mefellows:
Think I got to the bottom of it. The `matchHeader` method from nock was causing issues, updating to latest solved that problem. There was also a test that needed updating which I've also fixed.

mefellows:
Let me know if you need more from me on this one @mboudreau .

mboudreau:
Merge it?

mefellows:
Did you see my review?

mboudreau:
...I had not.  must have missed it.  Time to get fixin' :P

src/common/request.spec.ts
```diff
@@ -1,177 +1,69 @@
-/* tslint:disable:no-unused-expression no-empty */
+/* tslint:disable:no-unused-expression no-empty no-console */
```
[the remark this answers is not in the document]
mboudreau: :+1:

examples/typescript/test/get-dog.spec.ts
```diff
@@ -45,7 +45,7 @@ describe("The Dog API", () => {
         .given("I have a list of dogs")
         .uponReceiving("a request for all dogs")
         .withRequest({
-          method: "GET",
+          method: HTTPMethod.GET,
```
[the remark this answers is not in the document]
mboudreau: Preference mostly; I find it cleaner to write, you can use it in JS as well with HTTPMethod.GET, not sure TS and if you'd want to list them all possibilities, it's easy to do in both TS and JS for validation :)

package.json
```diff
@@ -5,17 +5,17 @@
...
     "coverage": "nyc report --reporter=text-lcov | coveralls",
     "deploy:prepare": "./scripts/create_npmrc_file.sh",
-    "dist": "npm run build && webpack --config ./config/webpack.web.config.js",
+    "dist": "npm run compile && webpack --config ./config/webpack.web.config.js",
     "jscpd": "jscpd -p src -r json -o jscpd.json",
     "lint": "standard && tslint -c tslint.json 'src/**/*.ts'",
     "postdist": "npm t",
     "posttest": "npm run test:karma",
     "predist": "npm run clean && npm run lint && npm run jscpd",
-    "prerelease": "npm i && rm package-lock.json",
```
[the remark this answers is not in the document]
mboudreau: is there really a needed? npm install is already done before a release from travis, package-lock.json is in the .gitignore, which won't be packaged in npm.

package.json
```diff
@@ -46,22 +46,22 @@
     "testing",
     "consumer driven testing"
   ],
-  "author": "Beth Skurrie <beth@bethesque.com> (https://github.com/bethesque)",
+  "author": "Matt Fellows <m@onegeek.com.au> (http://twitter.com/matthewfellows)",
```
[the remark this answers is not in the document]
mboudreau: Credit where credit is due buddy :)

src/pact.ts
mboudreau: ah yeah, sorry, missed that one.

src/pact.ts
mboudreau: yeah, wanted to add a TODO to it.  The issue here is that it seems that q promises are not interchangeable with regular Promises from a typing perspective.  I need to fix that so that we don't have this horrible issue of typing issues, but that's outside of the realm of this PR.  I really loathe the fact that JS promises don't have the `finally` just yet or any of the other niceties that q provides.

Anywho, this is the only way I could do it without having the TS compiler complaining about types not matching and having the promise actually wait for the server to delete since before it wasn't waiting for it. this will be temporary.

mboudreau:
@mefellows alright, fixed some things based on comments, or added some more information as to my decision making.

package.json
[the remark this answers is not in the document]
mefellows: I actually can't remember why this existed, perhaps it was before we added it to `.gitignore`. Was just wondering if it caused you grief or something.

mefellows:
Thanks @mboudreau, I think you still missed a couple of comments:

> My only concern is around the HTTPMethod change from type to enum. We should think carefully, as it will be a breaking change and i'm not so sure it's worth it. Happy to hear a case for it of course.

and

> OH - and lastly, could you please ensure the commits are in line with the CONTRIBUTING guidelines. It might be a task re-writing them all, so feel free to squash into a single commit. Depending on wheather or not it is a "BREAKING CHANGE" or a feature will effect the release process via npm run release.

mboudreau:
hm, right, It would be a breaking change for TS.  I could add it as a union type.

I'll squash the commits.

mboudreau:
Can't we just do the squash commit from the PR here? do a squash & merge?  I can do the actual merge so the message is appropriate to your guidelines.

mefellows:
Thanks for all of your hard work Michel - your code is in!

I had to also add a union type to `RequestOptions` to accept the `HTTPMethod` and `method` types, but over time we can drop the latter.

This made the change completely backwards compatible. 

`v5.6.1` will be on its way out shortly...

mbark:
Thanks for the good work! Our pact tests are now up and running! 🎉

mefellows:
Awesome! Thanks for your patience @mbark

PR #148, opened 2018-02-05
Require '@pact-foundation/pact' in inline example

mattcanty:
I think having this in the first example on the page would have cleared up a lot of problems from the start. An issue with me was going in quick running 'npm install pact-js' which would have put `pact` in my modules.

After installing the v5 @pact-foundation/pact I think my code references must have still been hitting v4.


mefellows:
Sorry mate, don't know how I missed this. Merging!

PR #162, opened 2018-03-19
Update provider option for local files

saschak094:
Update provider option pactFilesOrDirs with pactUrls


mefellows:
Hmm something is not right. As per the `pact-node` issue, if you take a look at the [file](https://github.com/pact-foundation/pact-node/blob/master/src/publisher.ts#L86), `pactFilesOrDirs` is the correct argument. 

Have you got an outdated version?

We also run each travis build against these examples, so it _should_ fail if the arguments are incorrect.

mefellows:
Ah, nevermind - I responded on mobile and hadn't seen this closed off. thx.

PR #164, opened 2018-03-20
Fix import in example

carwinz:
The example shows "like" being imported from 'pact'; it should be imported from "Matchers"


mefellows:
Thanks :)

Issue #173, opened 2018-04-19
Can't require() module in 6.0.0-alpha.8


TomRK1089:
## Software versions

* **OS**: Windows 7
* **Consumer Pact library**: Pact JS v6.0.0-alpha.8
* **Node Version**: `v8.9.4`

## Expected behaviour

I can run tests that import Pact like so:
```
const { Matchers, MessageConsumer, Message, synchronousBodyHandler } = require("@pact-foundation/pact");
```

## Actual behaviour

I get an error when invoking the test via Mocha:
```
Error: Cannot find module '../../package.json'
    at Function.Module._resolveFilename (module.js:538:15)
    at Function.Module._load (module.js:468:25)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (<redacted>\node_modules\@pact-foundation\pact\common\logger.js:15:11)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (<redacted>\node_modules\@pact-foundation\pact\pact.js:28:16)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (<redacted>\test\mq-consumer.test.js:1:134)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at <redacted>\node_modules\mocha\lib\mocha.js:253:27
    at Array.forEach (<anonymous>)
    at Mocha.loadFiles (<redacted>\node_modules\mocha\lib\mocha.js:250:14)
    at Mocha.run (<redacted>node_modules\mocha\lib\mocha.js:577:10)
    at Object.<anonymous> (<redacted>\node_modules\mocha\bin\_mocha:591:18)
[...]

PR #174, opened 2018-04-20
fix: Resolve package.json for version metadata regardless of nesting

TomRK1089:
The package.json file appears at a different relative location in the source tree versus the compiled package.
It should be resolved regardless of the level of nesting present in the calling module.

Relates to https://github.com/pact-foundation/pact-js/issues/173

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "consumer driven testing"
   ],
   "author": "Matt Fellows <m@onegeek.com.au> (http://twitter.com/matthewfellows)",
-  "contributors": [{
+  "contributors": [
+    {
       "name": "Tarcio Saraiva",
       "email": "tarcio@gmail.com",
       "url": "http://twitter.com/tarciosaraiva"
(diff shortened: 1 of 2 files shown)


mefellows:
Awesome - thanks for this! I'll push out a new alpha shortly...

mefellows:
hmm, merge build failed: https://travis-ci.org/pact-foundation/pact-js/builds/368919159?utm_source=email&utm_medium=notification

It's ok as it's not critical, but something to address when you/I get a moment.

mboudreau:
@mefellows Why would the path change on release?

TomRK1089:
Yes I just noticed the Jest tests were failing (somehow missed that when running locally). Trying to investigate now.

TomRK1089:
@mefellows trying again in https://github.com/pact-foundation/pact-js/pull/175 and seeing if Travis agrees with what I see in `npm test` locally now.

PR #178, opened 2018-05-09
fix(interaction): include response body if set to empty string

### LINKED ISSUE #177: Expecting an empty body Vs not caring about the body.
Hi! We've recently had an integration bug that should've been caught by contract testing but it wasn't.

## Software versions

* **OS**: e.g. Mac OSX 10.13.14
* **Consumer Pact library**: e.g. Pact JS v4.3.2
* **Provider Pact library**: -
* **Node Version**: v8.11.1

## Expected behavior

When the consumer is expecting the provider to return an empty body (i.e. content-length = 0)

```
myPactProvider.addInteraction({
       ...
       willRespondWith: {body: '', status: 200},
       ...
});
```
The generated pact file should contain the body property (set to empty string):

```
{
  "interactions": [
    {
       ...
       "response": {
         "body": "",
         "status": 200,
         "headers": {}
       }
    }
  ],
  ...
}
```

## Actual behaviour

The body property is removed altogether which means that it won't be validated:

```
{
  "interactions": [
    {
       ...
       "response": {
         "status": 200,
         "headers": {}
       }
    }
  ],
  ...
}
```

notice that other pact client libraries such as pact-jvm do care about this distinction (i.e. not defining a body vs setting it to empty).

I haven't checked the behavior for request bodies, but the bug might exist there as well.

I'm willing to raise a PR if you agree that this is a bug.

Thanks!

argos83:
Fixes #177 
The problem existed only for response bodies, but I also added tests for request bodies so as to prevent a regression.

### DIFF
--- a/src/dsl/interaction.spec.ts
+++ b/src/dsl/interaction.spec.ts
@@ -87,6 +87,26 @@ describe("Interaction", () => {
         expect(actual.request).to.have.keys("method", "path", "query", "headers", "body");
       });
     });
+
+    describe("request body", () => {
+      it("is included when an empty string is specified", () => {
+        const actual = new Interaction().withRequest({
+          body: "",
+          method: HTTPMethod.GET,
+          path: "/path",
+        }).json();
+        expect(actual.request).to.have.any.keys("body");
+      });
+
+      it("is not included when explicitly set to undefined", () => {
+        const actual = new Interaction().withRequest({
+          body: undefined,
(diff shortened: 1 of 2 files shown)


argos83:
🤔 the AppVeyor check failed, it seems unrelated to my changes though.

mefellows:
Thanks for this.

argos83:
> Thanks for this.

Thank you! you guys are always super responsive

PR #189, opened 2018-06-19
Making pact-js bower compatible

davidkgp:
Hello Team,

We as an organisation has adopted Pact as a mode of consumer contract driven testing and provider verification.Our UI build server uses bower dependency as an organisational configuration to download packages.Besides bower is an aspect of dependency management alongwith npm,so wanted to make this project bower compatible.As a part of this I have introduced a bower.json file and have registered this as a bower package in the bower registry.Kindly allow me to merge this change(single bower.json file ) to the project.While merging I would like to change the public git url to the pact foundation pact js url rather than my fork and re register the package in bower registry.


mboudreau:
I'm curious as to why you feel the need to use bower, when you can use npm or yarn exactly the same way?  Furthermore, [bower has been deprecated since October 2017](https://bower.io/blog/2017/how-to-migrate-away-from-bower/) and there hasn't been a new version since 2016.

Unless you can bring forward a very good case to support it, my vote is to not add more complexity to what we support.

davidkgp:
Currently my organisational build pipeline is based on fruitloops which is configured to use bower to download the dependencies.So this is a common framework which is not prone to change.I agree with ur point of bower being deprecated,but just adding bower support would mean the pact js library remains backward compatible and developers who are using bower can still continue adapting to pact way of working.

mboudreau:
[You mean this fruitloops](https://www.npmjs.com/package/fruit-loops)? The framework that hardly anyone uses (86 downloads a week?!) and hasn't been released in over 3 years?  It's not a common framework and I've personally never heard of it until now.  It's not prone to change because, well, nobody maintains it anymore.

Adding bower support _does not mean that "pact-js remains backwards compatible" at all_.  Bower is backwards compatible to Node 0.10, but the packages under it has no mandate to do the same, and frankly, why would we?  Our product is evolving and sometimes things will break, but we use proper semantic versioning standards so that things might break on a major version change only.  We also have a plethora of tests to make sure that any changes aren't breaking current functionality.

To further my point, pact-js requires pact-node to function, which in itself requires either npm or yarn since it downloads the binaries on the post-install hook.  None of this would work with bower, and your project will break.  I strongly recommend you move away from fruitloops and bower since they are both deprecated and unmaintained.

You've yet to make a compelling argument for your case, and I'm not going to spend several hours trying to figure out how to make pact-node compatible with bower, which will be all for naught since it's being deprecated.  I can guarantee that our product will be backwards compatible to node 4 and up, since all the tests run for every version of node from 4 until the latest.  I'm going to close this PR, unless a proper argument can be made or it's something that an abundant amount of our users want.

PR #197, opened 2018-07-06
example | verify multiple tests in one file by calling provider.verify() in afterEach block.

rashiagarwal:


mefellows:
Thanks @rashiagarwal, sorry I have no idea how I missed this. It's on my backlog to review (I need to fix that broken build first)

PR #204, opened 2018-07-26
close http server instance started by Message Provider Verifier

mhughesbv:
When running MessageProvider tests using Jest the test pack never quits. This has been narrowed down to a http server that is started when calling the verify() method but is never closed. A small
change to ensure that the server gets shut down correctly after the verification has been completed.


mefellows:
Thanks for this, my sincere apologies for not seeing it. It's on my list of stuff to review later today

PR #207, opened 2018-08-12
docs(README): update code example

lirantal:
- Added a comment for clarification about Pact version 
- Update code example with async call to use the callback pattern so it properly waits until completion


mefellows:
Thanks!

PR #211, opened 2018-09-19
feat(examples): add array bracket notation example to jest tests

lirantal:
## Changes
- Updated jest examples with an example reference on using query params with square bracket notation.
- Updated best practice to include the interaction in the test case
- Refactored the callback style to promises

## Open questions:
- Looks like when the tests run the pact file doesn't include the /cats route. Is that a bug in how the 'update' works or wrong setup on the jest tests ?
- Do you see any benefit in using the individual pact test wrapper


TimothyJones:
Thanks for this! I was looking at updating this example yesterday, and you've already fixed several of the issues.

Could you elaborate on what you mean by the the individual pact test wrapper?

examples/jest/__tests__/catAPI.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  let url = 'http://localhost'
+
+  describe("another works", () => {
+    it('returns a successful body', async () => {
+      // add expectations
+      const EXPECTED_BODY = [
+        { cat: 2 }, { cat: 3 }
+      ]
+
+      const interaction = {
```
TimothyJones: I personally think this reads better in the `beforeEach()` block, because it's the Given part of the Given/When/Then (this also would match the other example tests in the Jest folder)

examples/jest/__tests__/catAPI.spec.js
```diff
@@ -0,0 +1,53 @@
...
+        willRespondWith: {
+          status: 200,
+          headers: {
+            'Content-Type': 'application/json'
+          },
+          body: EXPECTED_BODY
+        }
+      }
+
+      await provider.addInteraction(interaction)
```
TimothyJones: Similarly to the comment above, I personally think that the `.then()` style is more readable, as it lets you separate the When from the Then.

examples/jest/__tests__/catAPI.spec.js
```diff
@@ -0,0 +1,53 @@
...
+      const response = await getMeCats({
+        url,
+        port
+      })
+
+      expect(response.headers['content-type']).toEqual('application/json')
+      expect(response.data).toEqual(EXPECTED_BODY)
+      expect(response.status).toEqual(200)
+
+      return provider.verify()
```
TimothyJones: It seems a bit unexpected for an `async` function to return a promise.

As above, I think the `.then()` style used in the other examples is better for separating the concepts in the test anyway.

However, if we decide to go with changing to `async/await`, I think we should change this function to also await `provider.verify()`, so as not to mix styles (and also update the other examples).

examples/jest/__tests__/catAPI.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+const path = require('path')
+const { Pact, Matchers } = require('../../../dist/pact')
+const getMeCats = require('../').getMeCats
+
+describe("Cat's API", () => {
+  let url = 'http://localhost'
+
+  describe("another works", () => {
+    it('returns a successful body', async () => {
```
TimothyJones: Ah - the use of `async` is failing the node 6 build. To provide the best experience for newcomers, we'd like to support node 6 without transpiling. Can you update the PR?

lirantal:
Thanks for the comments! I didn't have time to go through them and address but I'll get back to that in a couple of weeks.

TimothyJones:
Thanks for your work! Contributions are always appreciated!

lirantal:
@mefellows did we close because I was stalling the changes here?
If so, I'd be happy to address the comments and re-submit the PR.

mefellows:
I think I just meant to press Comment.

lirantal:
@TimothyJones could you review my changes and comments when you get a chance?

TimothyJones:
Sorry for the delay @lirantal (and thanks for the prompt!). I'll respond to your individual comments after posting this.

For changes, I can't see any extra commits from my side - did you push any?

Largely, though I think there are two main discussions here:

1) **Should we add an example that shows how to do array bracket notation?** I think the answer is "yes, absolutely". For this reason, I'd like to get this PR merged in.

2) **What is the best way to write a Pact test in Jest?** Since these tests are examples, it's good to have them written in a best-practice way (the current code in `master` is quite long, and could be more straightforward). I don't want to speak for you, but my guess is this is why you've included the refactoring in this PR (or, if not your motivation, it's certainly a nice advantage of the refactoring).

However, figuring out what a best-practice Jest test looks like is a bigger discussion - (in my own projects, I've been playing with not using a `testSetup.js`, and I've been considering working on a `pact-jest` module to reduce boilerplate). 

I don't want to delay accepting this PR because of a side discussion about the right test architecture - but I do want to have that side discussion, because I think it is interesting and valuable to the project. The better our examples are, the easier Pact is to pick up.

If you're keen, I think it would be valuable to work together on figuring out what the tests should look like (especially if we don't immediately share the same opinions).

Would you consider splitting this one into two PRs, one with the array brackets change written in the current style, and another that refactors the promises (and/or makes other changes towards best-practice tests)? What do you think? 

Please let me know if I have missed the mark or there is some reason that the array brackets tests rely on the promise refactoring.

examples/jest/__tests__/catAPI.spec.js
TimothyJones: I reckon it's worth keeping node 6 in CI at least until it's EOL'ed (and possibly for a bit after).

examples/jest/__tests__/catAPI.spec.js
TimothyJones: This is probably a personal style thing. To explain what I meant, my pact tests tend to look like this:

```
describe("A request for foo", () => {
    describe("In state bar", ()=> {
        beforeEach(() => {/* setup interaction here*/}
        it("generates the right response", ()=> 
            api.foo().then(/* assertions go here */);
        )
    });
    describe("In state baz", ()=> {
        beforeEach(() => {/* setup interaction here*/}
        it("generates the right response", ()=> 
            api.foo().then(/* assertions go here */);
        )
    });
});
```

In this way, each test has a single corresponding `beforeEach()` statement. I like this, as it means that the `describe`s and the `beforeEach` are the "Given", the API call is the "When", and the `.then(/* assertions */)` is the "Then".  I find this very easy to read, because it doesn't include "given" sections within the `it()`. You may prefer a different style - there are many good ways to write a test.

I brought this up in the review not because I want you to use my style, but because I think it's valuable to have all the examples written in the same style, to minimise cognitive load on the reader.

A Pact test has a lot more going on than a standard unit test, so a nice clear separation of the different components is valuable to the new reader (and could easily be a companion to a documentation article).

I don't feel strongly about what style we use, as long as the style fits the values of being easy for a new reader (from the perspective of understanding Pact), and is a good example of idomatic code in the example test framework + language.

However, if we are going to change the style of the examples, I do feel strongly that we should change all of them.

examples/jest/__tests__/catAPI.spec.js
TimothyJones: Apologies, I wasn't very clear there. Here's a better explanation of the first paragraph: 

`provider.verify()` returns a promise, and as you mention, making a function `async` effectively turns the function into a promise, meaning you can have more readable code that deals with values instead of promise chaining. For me, this means I expect functions that are marked `async` to look like this:

```
async () => {
    ...
    return <some value that isn't a promise>
}
```

instead of:

```
async () => {
    ...
    return <some promise>
}
```

Of course, this is fixed if we `await provider.verify()` (and not needed if we support node 6 by not using `await`).

examples/jest/__tests__/catAPI.spec.js
lirantal: That makes sense though, thanks for explaining that.
I'd be happy to take this in another PR though and address all the examples so we can unblock this one.

examples/jest/__tests__/catAPI.spec.js
lirantal: Regardless of using async function signature or not, if you need to return a value it doesn't matter if it's a primitive, complex or a promise itself.

What I'm trying to say is that for an async function that needs to return a value it is perfectly valid to have a `return <some promise>`, or to await the value and then return it. However just running the promise or awaiting the promise without returning it will not return the value explicitly.

So even if by your reference you `await provider.verify()` you will not return the value of `provider.verify()`.

lirantal:
Completely relate to your view of using examples as best practices at https://github.com/pact-foundation/pact-js/pull/211#issuecomment-438131849
Indeed I didn't open this PR to refactor for the best practices guidelines but instead to provide the array brackets example. I'm more than happy though to unblock this PR as you also stated so we can discuss best practices in a new PR.

Updated to reflect only the bracket notations.

mefellows:
As per the discussion, will merge this in and the example can be refactored later on.

PR #218, opened 2018-10-06
Add Windows build script with appveyor

### LINKED ISSUE #112: Create Windows (Appveyor) build
We currently suffer from a lack of visibility into Windows issues and reproduceability when there are issues.

We need to create a working Appveyor build as per other projects and make Windows a first class citizen!

mefellows:
Runs a similar build process to travis, except does not publish/release.

Fixes #112


scripts/build.sh
```diff
@@ -3,7 +3,7 @@
 npm run dist
 
 # Only run E2E examples on Node 6+
-if [[ "${TRAVIS_NODE_VERSION}" =~ ^(6|7|8|9|10) ]]; then
+if [[ "${TRAVIS_NODE_VERSION}" =~ ^(6|7|8|9)|1[0-9]+ ]]; then
```
TimothyJones: It would be neat if we could pull this logic outside of the two platform specific scripts - maybe with  [`check-node-version`](https://www.npmjs.com/package/check-node-version) - which also would mean the script keeps working when we get to node 20.

package.json
```diff
@@ -160,7 +159,9 @@
     "superagent": "^3.8.2",
     "ts-node": "^3.3.0",
     "tslint": "^5.8.0",
+    "tslint-config-prettier": "^1.15.0",
     "typescript": "^2.7.2",
-    "webpack": "^3.5.5"
+    "webpack": "^3.5.5",
+    "winston": "2.4.0"
```
TimothyJones: I'm probably missing something - what's the reason for adding more logging dependencies here?

appveyor.yml
```diff
@@ -0,0 +1,34 @@
...
+  - npm install
+
+# Post-install test scripts.
+test_script:
+  - node --version
+  - npm --version
+  - npm run dist
+  - powershell .\scripts\build.ps1
+
+# Don't actually build.
```
TimothyJones: I'm a little confused by the build steps - `dist` is run before `test` (which happens in `build.ps1`), and then there's a comment that says "don't actually build". Is there a way we can make this clearer?

package.json
mefellows: Yeah, that looks to be a mistake - good pickup.

appveyor.yml
mefellows: I kept it in line with the travis build, but agree it's confusing. `build.ps1` really should be `run_examples.ps1` sort of thing - that's all it does. I might refactor both after the merge.

As for the "Don't actually build" I'll rename to something like "Prevent default msbuild" as that is what Appveyor will try to do

mefellows: I've tidied this up now and think it's more readable - let me know what you think.

PR #233, opened 2018-10-17
fix(interaction-failure): clear interactions on any verify()

### LINKED ISSUE #231: Having a "no matching interaction" on a single test crashes the entire suite
## Software versions

* **Consumer Pact library**: `^5.9.1`
* **Node Version**: `8.11.3`

## Expected behaviour

Given I add an interaction
And I'm doing TDD red-green-refactor cycle
When my callApi module makes the HTTP request
And it doesn't match the defined interaction ("red" part of the cycle)
Then only the that test should crash
And the other tests should have consistent results

## Actual behaviour

All the tests that execute after the one that didn't match crashes.

## Steps to reproduce

```javascript
describe('callApi', () => {
  afterEach(() => provider.verify());

  it('a test that passes', () => {
    provider.addInteraction(interactionForEndpointA);
    callApi.endpointA();
  });

  it('a test that doesn\'t match the interaction', () => {
    provider.addInteraction(interactionForEndpointB);
    callApi.endpointB();
  });

  it('a test that should pass but doesn\'t, unless we comment out the previous test', () => {
    provider.addInteraction(interactionForEndpointC);
    callApi.endpointC();
  });
})
```

## Relevant log files

```
ERROR -- : No matching interaction found for PUT /my/endpointB
ERROR -- : Interaction diffs for that route:
ERROR -- : Diff with interaction: "some interaction" given "some state"

Diff
--------------------------------------
Key: - is expected 
     + is actual 
Matching keys and values are not shown

 {
-  "body": {
-    "myProp": "myValue"
-  }
+  "body": "[object Object]"
 }

Description of differences
--------------------------------------
* Expected a Hash but got a String ("[object Object]") at $.body

WARN -- : Verifying - actual interactions do not match expected interactions. 
Incorrect requests:
	PUT /my/endpointB (request body did not match)

WARN -- : Incorrect requests:
	PUT /my/endpointB (request body did not match)


...... now it will register interactions for the next tests
...... and output that it found a response for the call in the next test
...... however, the verification is done against the PUT call above

INFO -- : Registered expected interaction GET /my/endpointC
DEBUG -- : {
...... the interaction definition
}
INFO -- : Received request GET /my/endpointC
DEBUG -- : {
....... the request
}
INFO -- : Found matching response for GET /my/endpointC
DEBUG -- : {
....... the response 
}
WARN -- : Verifying - actual interactions do not match expected interactions. 
Incorrect requests:
	PUT /my/endpointB (request body did not match)

WARN -- : Incorrect requests:
	PUT /my/endpointB (request body did not match)

........ I would have expected it would verify with the NEXT registered interaction, 
........ not with the one that already failed.
........ Also, the same thing happens for the rest of the tests
........ They all fail cause it's trying to verify with the failed interaction
```

### Update
I noticed the tests that pass have this line in the logs:
```
INFO -- : Cleared interactions
```

This line is not present after a verification that failed. Maybe that's the problem?

mefellows:
All successful/failed calls to pact.verify() should clear
interactions so that subsequent tests don't det out of whack.

NOTE: Promise.finally() would have been nicer, note for
upgrade to latest TS in next major release.

Fixes #231

### DIFF
--- a/src/pact.spec.ts
+++ b/src/pact.spec.ts
@@ -245,7 +245,7 @@ describe("Pact", () => {
         const verifyPromise = b.verify();
         return Promise.all([
           expect(verifyPromise).to.eventually.be.rejectedWith(Error),
-          verifyPromise.catch(() => expect(removeInteractionsStub).to.callCount(0)),
+          verifyPromise.catch(() => expect(removeInteractionsStub).to.callCount(1)),
         ]);
       });
     });
--- a/src/pact.ts
+++ b/src/pact.ts
@@ -127,6 +127,7 @@ export class Pact {
     return this.mockService.verify()
       .then(() => this.mockService.removeInteractions())
       .catch((e: any) => {
+
         // Properly format the error
         /* tslint:disable: no-console */
         console.error("");
@@ -134,8 +135,10 @@ export class Pact {
         console.error(clc.red(e));
         /* tslint:enable: */
 
-        throw new Error("Pact verification failed - expected interactions did not match actual.");
-      });
+        return this.mockService.removeInteractions().then(() => {
+          throw new Error("Pact verification failed - expected interactions did not match actual.");
+        })
+      })
   }
 
   /**


TimothyJones:
Nice!

PR #239, opened 2018-10-28
updated explanation of providerStatesSetupUrl

thombergs:
I tried to clarify the description of `providerStatesSetupUrl` a little since I had assumed that it's mandatory when provider states have been defined in the pact


TimothyJones:
Nice work, thanks!

PR #244, opened 2018-11-05
fix(vulnerabilities): upgrade version of pact-node

TimothyJones:
Including the current version of `pact-js` in a github project results in security warnings for contributors. (See [here](https://github.com/pact-foundation/pact-node/pull/131) for details).

This PR fixes those warnings by upgrading pact-node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks Tim!

PR #245, opened 2018-11-06
De-duplicating link to Typescript example.

AlexanderPruss:
Just a tiny typo fix I noticed while reading through the docs.


TimothyJones:
Good catch, thanks!

PR #246, opened 2018-11-07
fix(logger): apply logLevel value passed via constructor

vgrigoruk:
Currently, `logLevel` is passed only downstream to pact-node and not respected by a local logger.

Steps to reproduce:

```
const path = require('path');
const { Pact } = require('@pact-foundation/pact');

const provider = new Pact({
  port: 9999,
  host: '0.0.0.0',
  cors: true,
  log: path.resolve(process.cwd(), 'pact.log'),
  logLevel: 'fatal',
  consumer: 'consumer',
  provider: 'provider'
});
```
Expected: no logs in STDOUT
Actual: INFO logs in stdout.

This PR fixes the issue.

### DIFF
--- a/src/pact.ts
+++ b/src/pact.ts
@@ -13,6 +13,7 @@ import { Interaction, InteractionObject } from "./dsl/interaction";
 import { isEmpty } from "lodash";
 import { isPortAvailable } from "./common/net";
 import logger from "./common/logger";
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks @vgrigoruk!

vgrigoruk:
Thanks for merging, @mefellows. I was wondering why CI failed, but the failure was not related to my changes.

PR #252, opened 2018-11-22
iso8601DateTimeWithMillis: allow variable length fractional precision

fewstera:
I was attempting to use the iso8601DateTimeWithMillis matcher on a field like `2018-11-21T14:44:31.537357Z`, which was being returned from a Go service, but the match was failing.

Currently, iso8601DateTimeWithMillis only allows the precision of the ISO8601 fractional seconds to 3 decimal places, this PR allows any number of decimal places.

I'm slightly unsure of this PR as the function is named `...WithMillis`, but this PR will allow precisions greater than a millisecond. I didn't want to break backwards compatibility, so renaming the function wasn't an option, but maybe it could live as a separate function like `iso8601DateTimeWithFractionalSecond`. Be interested to hear your thoughts.

### DIFF
--- a/src/dsl/matchers.spec.ts
+++ b/src/dsl/matchers.spec.ts
@@ -505,6 +505,7 @@ describe("Matcher", () => {
       describe("when given a valid iso8601DateTimeWithMillis", () => {
         it("should not fail", () => {
           expect(iso8601DateTimeWithMillis("2015-08-06T16:53:10.123+01:00")).to.be.an("object");
+          expect(iso8601DateTimeWithMillis("2015-08-06T16:53:10.537357Z")).to.be.an("object");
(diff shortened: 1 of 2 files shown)


mefellows:
Thanks for this. I'll check against the other implementations to ensure consistency where possible, but failing that, I don't see a big issue in supporting a new matcher if it makes sense.

I'll take a look next week, hopefully.

mefellows:
Looking at the Java implementation, it's not totally consistent anyway. I don't think it's an issue because it won't cause compatibility issues (the matcher is always going to be contained in the pact file anyway).

Happy to bring this in, but I'll modify it so that it has _at least_ 3 points of precision.

PR #266, opened 2019-02-09
fix(port-check): make port check more resilient. Fixes #49

### LINKED ISSUE #49: Tests timeout before pact daemon starts
I've started experiencing an issue where my mocha tests time out before the pact daemon starts.
```
[2017-05-12T00:19:06.959Z]  INFO: pact-node@4.9.1/36673 on Machine: Creating Pact with PID: 36674
    1) "before all" hook


Pact verification failed!
Error: connect ECONNREFUSED 127.0.0.1:6666
(node:36673) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Pact verification failed - expected interactions did not match actual.
(node:36673) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:36673) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: connect ECONNREFUSED 127.0.0.1:6666

  0 passing (2s)
  1 failing

  1) Contracts Pact "before all" hook:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

The default two second timeout is too short, I've resolved this by increasing the timeout on mocha to make the tests pass.
```
mocha consumer.js --timeout 10000
```

The weirdness is:

1. This has just started happening, nothing changed in terms of dependency versions
2. It was not clear to me what the problem was but reading the output from mocha, I had to probe around to figure this out.

mefellows:


src/common/net.ts
```diff
@@ -5,16 +5,31 @@
  */
 
 import * as net from "net";
+import { Promise as bluebird } from "bluebird";
 
-const isPortAvailable = (port: number, host: string): Promise<void> => {
+const isPortAvailable = (port: number, host: string): Promise<void> =>
+  Promise.resolve(bluebird.each([host, "127.0.0.1", "localhost", "0.0.0.0"], h =>
```
TimothyJones: Is this the right set of addresses? 0.0.0.0 means all addresses on the local machine, which might not always be right (and it’s also possible that there’d be cases where we don’t want it listening on the loopback, too?)

src/common/net.ts
```diff
@@ -5,16 +5,31 @@
...
   return new Promise((resolve, reject) => {
-    const server: any = net.createServer()
+    const server: any = net
+      .createServer()
       .listen({ port, host, exclusive: true })
-      .on("error", (e: any) => (e.code === "EADDRINUSE" ? reject(new Error(`Port ${port} is unavailable`)) : reject(e)))
-      .on("listening", () => server.once("close", () => resolve()).close());
+      .on("error", (e: any) => {
+        if (e.code === "EADDRINUSE") {
+          reject(new Error(`Port ${port} is unavailable`))
```
TimothyJones: I think this should contain the host too, given that we’re expanding the hosts above.

src/common/net.spec.ts
```diff
@@ -21,22 +21,14 @@ describe("Net", () => {
 
     context("when the port is available", () => {
       it("should return a fulfilled promise", () => {
-        expect(isPortAvailable(port, host)).to.eventually.be.fulfilled;
+        expect(isPortAvailable(port + 1, host)).to.eventually.be.fulfilled;
```
TimothyJones: I’m not sure I followed the +1 here. Is 4568 available, but 4567 isn’t? Is that specified soemwhere that I missed?

src/common/net.ts
mefellows: I ummed and ah'd about it. Doing some testing on my Mac, I found that even when I listened for a connection on the loopback, the port check would pass but the mock service would fail (with the EADDRINUSE. I'm not sure if there is something in the mock server that doesn't like it.

I felt given that it's better to fail fast (and eagerly) than otherwise, and that given it's a testing tool the environment should be more flexible.

Unless I'm blind to a class of errors?

src/common/net.spec.ts
mefellows: I wasn't able to fully track down the sequence of events, but it seems as if the OS doesn't fully release the port in the check when the next assertion is running - and thus the port isn't available.

...and then I realised my mistake - not _returning_ the `expect`. 🤦‍♂️

Fixed.

PR #271, opened 2019-02-21
fix(matchers): remove hardcoded true value from a boolean matcher

vgrigoruk:
Currently, a value passed as an argument to `boolean()` matcher is ignored, and the value that is returned by mockserver is always `true`.
This PR to fixes this:
- `boolean()`, `boolean(true)` -> `true`
- `boolean(false)`, -> `false`


mefellows:
Good one, thanks.

Issue #270, opened 2019-02-21
Proposal: pact-web provider.addInteraction to also allow Interaction instance


eijiew:
Currently pact and pact-web `provider.addInteraction` differs by pact accepting `InteractionObject | Interaction` as the argument whereas pact-web only accept `InteractionObject`

Could I propose for pact-web `provider.addInteraction` to also take in `Interaction`? So that it will be more consistent with pact implementation. Another advantage will be allowing pact-web to support interaction such as `GraphQLInteraction`

mefellows:
Thanks, this is a good idea. Happy for you to PR or for somebody to pick this up if they have time.

eijiew:
Thank you. I will proceed to close this Issue 😃

mefellows:
No worries. I'll release this in the next few days

PR #274, opened 2019-02-22
feat: pact-web provider.addInteraction to allow Interaction instance

eijiew:
pact-web `provider.addInteraction` to accept `Interaction` as argument

relating to #270


mefellows:
Like like build failed, if you wouldn't mind fixing first before I take a look?

mefellows:
Thanks! 👍

Issue #281, opened 2019-03-12
Cannot run provider verification against HTTPS endpoint that has CA, if called from a http endpoint


YOU54F:
For the example provided in https://github.com/pact-foundation/pact-js/issues/280, we were attempting to hit providerBaseUrl with a HTTPS end-point, self-signed without a CA. This is now working with the v8.0.4 release and the introduction of `validateSSL` flag.

In our work, we were attempting to hit providerBaseUrl with a HTTPS end-point, self-signed but with a CA, so we need `validateSSL` flag to set to true. I was unable to hit it and was getting the following error 

```
Error: write EPROTO 4515927488:error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:../deps/openssl/openssl/ssl/record/rec_layer_s3.c:1407:SSL alert number 40
```

I had a google and found a few similar issues but this thread had the answer that has resolved it

https://stackoverflow.com/questions/14262986/node-js-hostname-ip-doesnt-match-certificates-altnames

specifically this

```
n case you are using http-proxy package this happens when your server is HTTP (for example localhost) and target is HTTPS. To fix this issue set changeOrigin to true.

const proxy = httpProxy.createProxyServer();

proxy.web(req, res, {
  changeOrigin: true,
  target: https://example.com:3000,
});
In case your server is HTTPS and target one is HTTPS as well, you should include SSL certificate

httpProxy.createServer({
  ssl: {
    key: fs.readFileSync('valid-ssl-key.pem', 'utf8'),
    cert: fs.readFileSync('valid-ssl-cert.pem', 'utf8')
  },
  target: 'https://example.com:3000',
  secure: true
}).listen(443);
```

I updated verifier.js (in pact-js that is sitting in node_modules) to include the `changeOrigin` param and my requests are working

```
    // Get the Express app that will run on the HTTP Proxy
    Verifier.prototype.createProxy = function () {
        var _this = this;
        var app = express();
        var proxy = new HttpProxy();
        app.use(this.stateSetupPath, bodyParser.json());
        app.use(this.stateSetupPath, bodyParser.urlencoded({ extended: true }));
        // Allow for request filtering
        if (this.config.requestFilter !== undefined) {
            app.use(this.config.requestFilter);
        }
        // Setup provider state handler
        app.post(this.stateSetupPath, this.createProxyStateHandler());
        // Proxy server will respond to Verifier process
        app.all("/*", function (req, res) {
[...]

PR #282, opened 2019-03-12
add changeOrigin flag for http-proxy and update readme

YOU54F:
To address #281

### DIFF
--- a/src/dsl/verifier.ts
+++ b/src/dsl/verifier.ts
@@ -157,7 +157,7 @@ export class Verifier {
     app.all("/*", (req, res) => {
       logger.debug("Proxing", req.path)
       proxy.web(req, res, {
-        changeOrigin: this.config.changeOrigin === false,
+        changeOrigin: this.config.changeOrigin === true,
         secure: this.config.validateSSL === true,
         target: this.config.providerBaseUrl,
       })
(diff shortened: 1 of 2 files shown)


src/dsl/verifier.ts
```diff
@@ -156,6 +157,7 @@ export class Verifier {
     app.all("/*", (req, res) => {
       logger.debug("Proxing", req.path)
       proxy.web(req, res, {
+        changeOrigin: this.config.changeOrigin === false,
```
mefellows: Is this correct? Doesn't this mean if you set `changeOrigin` to `false` it would set it to `true`?

YOU54F: Will pull in the project on this branch now and just test this out, you are probably correct , apologies!

mefellows: No worries. I think you could copy the check for `secure`.

YOU54F: Cheers, just checked it out and it was set the wrong way around. Push up the change now.

I can't get my head around why the need for the `=== true` check.

I get the same behaviour without the check, if I pass in true, false as booleans, or don't provide, as lines 214-216 will set it to false if undefined.

If I just amend the js file to avoid typescript checking as pass in a string  in opts such as `  changeOrigin: "cheese",` rather than the boolean specified in the typescript interface, http-proxy seems to accept the fact that changeOrigin is provided irrespective of value.

PR #290, opened 2019-03-30
chore(docs): add note on Angular to troubleshooting section in README

eganjs:
I spent a lot of time attempting to integrate pact with an Angular project and wanted to share where I lost a lot of time so others wouldn't have to.

I'm not sure if you accept pull requests like this, but I thought I'd put it up just in case. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ Read [Getting started with Pact] for more information for beginners.
     - [Re-run specific verification failures](#re-run-specific-verification-failures)
     - [Timeout](#timeout)
     - [Note on Jest](#note-on-jest)
+    - [Note on Angular](#note-on-angular)
     - [Debugging](#debugging)
   - [Contributing](#contributing)
   - [Contact](#contact)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -971,6 +972,21 @@ Jest also runs tests in parallel by default, which can be problematic with Pact
 See [this issue](https://github.com/pact-foundation/pact-js/issues/10) for background,
 and the Jest [example](https://github.com/pact-foundation/pact-js/blob/master/examples/jest/package.json#L10-L12) for a working example.
 
+### Note on Angular
+
+Angular's HttpClient filters out many headers from the response object, this may cause issues when validating a response in tests.
+
+You'll need to add the additional header `Access-Control-Expose-Headers`, this will allow specified headers to be passed to the response object. This can be done in one of two ways, using `customProviderHeaders` as described [here](#modify-requests-prior-to-verification-request-filters) or by declaring the header in your contract:
+
+```js
```
mefellows: Is this bit supposed to be on the response  object or the request? (I'm assuming response).

It might be clearer to put a bit more context around this for others, although I'll rely on your experience to suggest whether or not this will be immediately obvious to other Angular users .

eganjs: It is supposed to be the response, I've added a bit more to try and make that clear.

mefellows:
Thanks. I'm also confused as to how `customProviderHeaders` would fix this issue, as this property can only work during provider side testing (and thus the property is not relevant during consumer tests).

eganjs:
> 
> 
> Thanks. I'm also confused as to how `customProviderHeaders` would fix this issue, as this property can only work during provider side testing (and thus the property is not relevant during consumer tests).

Looks like I've misunderstood that, I admit I was presumptuous there, I'll remove that part.

PR #295, opened 2019-04-15
fix(verifier): do not ignore providerStateSetupUrl passed to Verifier constructor 

vgrigoruk:
Currently, `providerStateSetupUrl` passed to `Verifier` constructor is always ignored, as `opts.providerStateSetupUrl` always overrides it. `providerStateSetupUrl` was deprecated in v8, but I still expect that it should to be possible to set it. 
This allows us to upgrade pact to v8 on our side without a full re-write of our custom provider state handling logic (as v7 is not maintained anymore).

This PR fixes described issue, so:
`providerStateSetupUrl` setting passed to Verifier constructor takes precedence over internal proxy server's `providerStateSetupUrl`.

### DIFF
--- a/src/dsl/verifier.ts
+++ b/src/dsl/verifier.ts
@@ -106,13 +106,11 @@ export class Verifier {
   private runProviderVerification() {
     return (server: http.Server) => {
       const opts = {
+        providerStatesSetupUrl: `${this.address}:${server.address().port}${
+          this.stateSetupPath
+        }`,
         ...omit(this.config, "handlers"),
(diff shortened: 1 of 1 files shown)


vgrigoruk:
Any feedback on this, @mefellows?

mefellows:
Hi @vgrigoruk. Apologies, it's been a looong weekend here in Aus. I'll take a look this week, I promise :)

mefellows:
Yep happy to support this change, sneaky but effective.

The side effect of this is that any state handlers attached in the new way will be useless. At this stage I'm reticent to put a log/notice in there. Let's see how it goes as is.

PR #313, opened 2019-06-15
feat(matching): allow integers to be treated as floats in verification

mefellows:
See  See https://github.com/pact-foundation/pact-ruby/issues/191 some background on this.

Allows integers to be treated as floats (but not the other way around) in provider verification.

This is currently intended to be part of a discussion, and should not yet be merged.

The proof is in the e2e example that now uses floating point examples in the consumer test but has a mix of integers and floats in the provider response.

### DIFF
--- a/examples/e2e/data/animalData.json
+++ b/examples/e2e/data/animalData.json
@@ -1,8 +1,9 @@
-[{
+[
+  {
     "first_name": "Billy",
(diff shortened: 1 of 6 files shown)


mefellows:
No longer needed - fixed in upstream Ruby.

Issue #299, opened 2019-05-15
Bluebird is a dependency, not a devDependency


mgabeler-lee-6rs:
## Software versions

* **OS**: Linux
* **Consumer Pact library**: Pact JS v8.2.2
* **Provider Pact library**: N/A
* **Node Version**: `10.15.3`

## Expected behaviour

`@pact-foundation/pact` package lists its dependencies correctly so that it works without relying on transitive dependencies which tools like `pnpm` block

## Actual behaviour

`src/common/net.ts` imports `bluebird` unconditionally, thus making it a "real" dependency and causing the `require` to fail when used in a `pnpm` environment.

Observe: 
* https://github.com/pact-foundation/pact-js/blob/master/src/common/net.ts#L8
* https://github.com/pact-foundation/pact-js/blob/master/package.json#L124

## Steps to reproduce

Setup a `rush` repo using `pnpm`.  Add `@pact-foundation/pact` as a dependency and have one of your (test) files try to load it.

## Relevant log files

Trimmed output of the above repro case:
```
Error: Cannot find module 'bluebird'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:582:15)
    at Function.Module._load (internal/modules/cjs/loader.js:508:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/home/user/src/rush-template/common/temp/node_modules/.registry.npmjs.org/@pact-foundation/pact/8.2.2/node_modules/@pact-foundation/src/common/net.ts:8:1)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
```
(Note: sourcemaps with bad relative paths are shipped, so the above `.../@pact-foundation/src/common/net.ts` path really means `.../@pact-foundation/pact/common/net.js`

mefellows:
Thanks @mgabeler-lee-6rs, looks to be a mistake. I'll get it updated.

mgabeler-lee-6rs:
FYI `@types` are always a dev dependency -- they're just needed for running `tsc`

YOU54F:
@mefellows @mgabeler-lee-6rs 

Hey Matt & Matthew,

This has broken one of my builds. Is the expectation that we need to include the @types/express in our consuming project?

```
#!/bin/bash -eo pipefail
yarn run build
yarn run v1.15.2
$ rimraf dist && tsc
[...]

PR #314, opened 2019-06-19
chore(deps): Move @types/* to devDependencies

vgrigoruk:
@types/* packages should be in devDependencies instead of runtime dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,9 +77,6 @@
   "homepage": "http://docs.pact.io/documentation/javascript.html",
   "dependencies": {
     "@pact-foundation/pact-node": "^8.6.0",
-    "@types/express": "^4.11.1",
-    "@types/bluebird": "^3.5.20",
-    "@types/bunyan": "^1.8.3",
     "bluebird": "~3.5.1",
     "body-parser": "^1.18.2",
     "bunyan": "^1.8.12",
@@ -102,9 +99,12 @@
   },
   "devDependencies": {
     "@pact-foundation/karma-pact": "~2.1.9",
+    "@types/bluebird": "^3.5.20",
+    "@types/bunyan": "^1.8.3",
     "@types/chai": "^4.0.3",
     "@types/chai-as-promised": "0.0.31",
     "@types/cli-color": "^0.3.29",
+    "@types/express": "^4.11.1",
     "@types/form-data": "^2.2.1",
     "@types/lodash": "^4.14.73",
     "@types/lodash.isnil": "^4.0.3",


mefellows:
I thought the same, but actually these are needed for some typescript configuratuons (see https://github.com/pact-foundation/pact-js/issues/299#issuecomment-492627721 for background).

Unless there is something we're missing?

vgrigoruk:
did see that one, @mefellows. Looks like this ticket needs to be closed then.
It is just a bit annoying to pull @types/* in JS project, where we don't need them.

TimothyJones:
I agree. This isn’t pact-ts (even if it is largely written in typescript).

Should the types be extracted to a PR to @types?

mefellows:
Good idea @TimothyJones. I agree, although we're talking about a minor cost (and probably zero inconvenience) it's not ideal having the types exposed.

@vgrigoruk would you be willing to look into that? I've not done it before, but presumably it isn't super difficult.

YOU54F:
what is the issue with the additional types files being pulled in? the additional space required should be minimal, and unless you are requiring pact-js as a dependency in your production packages, it would get bundled with your project if you build with webpack or similar.

I think because express is imported into the type definition file for the verifier, it will cause typescript users to break, and may cause issues for users with yarn/pnpm package managers.

I wonder if a better place for them is devDeps and peerDeps and a user can choose to install, if they wish.

https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies
https://github.com/LionC/express-basic-auth/issues/4

I don't transpile my verifier code into javascript, because it's not production code, and it run it with ts-node, so it uses the verifier.d.ts typing file supplied with the compiled pact-js, and therefore requires @types/express.

There seems to be no definitive answer on whether the consumers of a package, should pull in the @types if they need them (the error is pretty informative) and I have no issue in pulling them in my project, rather than the types being shipped for all.

also to note, even if these 3rd party types aren't included, every consumer of pact-js will pull in type definitions when pulling down the pact js module from npm, as the tsconfig has `declartion:true` which is good for me, and we write our pact tests in typescript, but unneccessary potentially for plain js users.

For now, I am happy with them being moved to devDeps, pulling in `@types/express` into my work and home projects, and we can see if this had a wider impact on other ts consumers of pact-js.

I wish there appeared to be official guidance from microsoft/definittlyTyped with regards to this, but I am yet to find it!

YOU54F:
submitting a PR to defTyped - https://github.com/DefinitelyTyped/DefinitelyTyped  seems a sensible suggestion from @TimothyJones but I am not sure how easy it is to keep it up to date, when you make changes to your interfaces, where are shipping it directly with your codebase means that it's always in sync.

TimothyJones:
Hm, that’s a good point about release cycle. Perhaps there’s a case for @pact-foundation/types?

I don’t actually use typescript myself, so I’m not super familiar with the ecosystem. 

I don’t think it’s high priority, but it seems like this dependency on another library’s types shouldn’t be something we depend on in a (published as) JS project- especially if it’s causing issues for people.

Providing types for Pact-js definitely provides value for users, I just don’t think it should be the responsibility of a repository published in JS.

How do other big JS projects provide their types?

mefellows:
These are all really good points (and very much thanks!), but I'm wondering if a PR is now the place to manage this. Carry on for now, as I can't move it to an issue. I can see both sides and would like to get to a nice resolution - some weekend reading for me coming up!

Issue #179, opened 2018-05-09
Examples are too contrived to be useful


mefellows:
The Dog API examples are useful, however as they are so simplistic the true essence of contract testing is not encouraged, and instead we encourage "API Testing" by asserting the response body, instead of how collaborating components behave.

Ideally, we have a single test case that is used consistently through all examples (perhaps sharing the domain model and common objects). My current thinking is that the E2E example should be the starting point.

TimothyJones:
Similarly, I think there could be some refactoring around the examples:

* They currently reference the parent folder to get `pact`, which is a barrier to being copied and pasted into a project (a common way to get started with a new framework)
* We don't have a clear separation between provider examples and consumer examples
* Some examples clear the pact folder each run, and some don't

plus other things I haven't noticed yet.

rahulballal:
For the problem that `pact` intends to solve. It would be nice if we can take a leaf out of this project https://github.com/gothinkster/realworld

Also, it would be nice if we get the meaty examples done first
1. REST api server - REST api client
2. Working with most popular test frameworks and retire the obsolete ones. Eg. Very few people use ava and tape AFAIK.
3. Replace the `examples/e2e` with a full docker compose setup. It is move prevalent.

All feedback is welcome ☮️

TimothyJones:
That's a well organised project! I like the way it's easy to understand their many repositories.

I think most of the examples will be a REST API server or client - especially when it's just the test runner that is changing - but we do have a bit of conceptual difficulty in that the message pact example is a very different way of connecting, and the end-to-end example is more of an example test infrastructure. Hmm.

I absolutely agree with focussing on the most popular test frameworks (npm-stat shows `mocha` and `jest` downloads absolutely dwarf `ava` - although I suppose ava uptake does seem to be growing). I don't think there's any harm in keeping uncommon frameworks around though, unless they're causing maintenance overhead.

I agree that `docker-compose` instead of Vagrant seems like an improvement. I'm not sure that end-2-end is the right name for that example - maybe "full pact workflow"?

mefellows:
@rahulballal I like the idea of a meaty project like realworld, I'll look into it a bit further.

> Working with most popular test frameworks and retire the obsolete ones. Eg. Very few people use ava and tape AFAIK.

[...]

PR #318, opened 2019-06-21
Update examples to run independently

mefellows:
Currently, all of the examples run with the implicit parent package scope, as well as requiring a compiled build of pact-js to run (i.e. a `npm run dist` in the root directory).

This makes it difficult for newcomers to navigate to an example and run independently, or to use as a basic for a project because the dependencies are not made clear.

This change does the following:

* Moves the karma examples into the examples, because it just makes sense to do so
* Uses the _exported_ versions of the pact packages (as one would do in real life)
* Adds all required packages to each example to enable independently installing
* Updates the travis build to use `npm link` so that all examples in the repository are still tested against the _current_  build of Pact JS / Pact Web, to prevent bugs/regressions


mefellows:
Thanks.

1. - Yes, probably a good idea (or we can just let the updates come through and fix them?)
2. There's not a lot of consistency in the examples, I'd love to see a consistent use case used throughout (where appropriate). See #179 for some ramblings on that.
3. Thanks, that would be appreciated.

PR #335, opened 2019-07-17
feat(matching): add string type matcher

### LINKED ISSUE #323: Feature request – create matchers for all JS data types
## Expected behaviour

It would be nice to add the following data types matchers:

– string
– symbol
– null
– undefined
– null
– object

## Actual behaviour

We have support only for:
– boolean
– number

abramenal:
Closes #323

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,7 @@ Often times, you find yourself having to re-write regular expressions for common
 | method                      | description                                                                                                                 |
 | --------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
 | `boolean`                   | Match a boolean value (using equality)                                                                                      |
+| `string`                    | Match a string value                                                                                                        |
 | `integer`                   | Will match all numbers that are integers (both ints and longs)                                                              |
 | `decimal`                   | Will match all real numbers (floating point and decimal)                                                                    |
 | `hexadecimal`               | Will match all hexadecimal encoded strings                                                                                  |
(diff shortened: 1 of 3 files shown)


src/dsl/matchers.ts
```diff
@@ -253,6 +253,13 @@ export function boolean(value: boolean = true) {
   return somethingLike<boolean>(value)
 }
 
+/**
+ * String Matcher.
+ */
+export function string(value: string = '') {
+  return somethingLike<string>(value);
```
mefellows: Can you please use a non-empty string here? You can choose something interesting, otherwise something like `"example"` will do.

README.md
```diff
@@ -609,6 +609,7 @@ Often times, you find yourself having to re-write regular expressions for common
 | method                      | description                                                                                                                 |
 | --------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
 | `boolean`                   | Match a boolean value (using equality)                                                                                      |
+| `string`                    | Match a string value (using equality)                                                                                       |
```
mefellows: This won't match on equality, but rather type. I'd suggest leaving as "Match a string value"

src/dsl/matchers.ts
abramenal: Done

README.md
abramenal: You're right. Done!

abramenal:
@mefellows updated, please take a look 😈

abramenal:
@mefellows

mefellows:
Apologies, have been AWOL prepping for a talk this week. Merging.

PR #339, opened 2019-07-30
chore(deps): Typescript example - maintenance - versions, warnings, speedup

andreasmarkussen:
# Minor updates for the TypeScript example:

 * ~~Mocha Maintainance - removed deprecation warning (--compilers)
    https://github.com/mochajs/mocha/wiki/compilers-deprecation~~ moved to PR 340
 * Updated dependencies (bumped versions in package.json)
 * Added a 'run script' that runs Mocha directly - 20% faster but without code-coverage

PR #340, opened 2019-07-31
chore(config): TypeScript Example - mocha.opts removing --compilers since it is depreciated

andreasmarkussen:
Mocha Maintainance - removed deprecation warning (--compilers)
https://github.com/mochajs/mocha/wiki/compilers-deprecation

This is a was of [PR 339](https://github.com/pact-foundation/pact-js/pull/339) but PR399 currently fails on older node versions.

### DIFF
--- a/examples/graphql/mocha.opts
+++ b/examples/graphql/mocha.opts
@@ -1,6 +1,6 @@
 --bail
---require ./test/helper.ts
 --require ts-node/register
+--require ./test/helper.ts
 --require source-map-support/register
 --full-trace
 --reporter spec
--- a/examples/typescript/test/mocha.opts
+++ b/examples/typescript/test/mocha.opts
@@ -1,7 +1,6 @@
 --bail
---compilers ts:ts-node/register
---require ./test/helper.ts
 --require ts-node/register
+--require ./test/helper.ts
 --require source-map-support/register
 --full-trace
 --reporter spec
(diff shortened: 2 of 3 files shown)


andreasmarkussen:
The PR fails on the GraphQL example that I did not touch, so this seems to be a flaky test. 
```
[2019-07-31T19:52:10.624Z]  WARN: pact@8.2.6/2292 on APPVYR-WIN: No state handler found for "null", ignorning
    1) validates the expectations of Matching Service
  0 passing (10s)
  1 failing
  1) Pact Verification
       validates the expectations of Matching Service:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\projects\pact-js\examples\graphql\provider.spec.ts)
```
In other cases it runs in about 9,4 seconds, so a little deviation is not unlikely in my mind? 

Should I make another PR that updates the timeout to e.g. 20 secs? 
And wait for that, before merging this?

andreasmarkussen:
@mboudreau  - do you have any idea what to do with the "Manifest not found"

TimothyJones:
We can safely ignore manifest not found- it’s a quirk of how synk remembers projects.

Thanks for updating this- I’ll take a look in more detail later today.

TimothyJones:
Very much appreciated, thank you!

PR #349, opened 2019-09-02
fix(graph-ql): Allow withOperation to accept null

shikhaFNZC:
Issue:
I am using apollo client with http link. I am not adding any operation name in the query/mutation so operation name is passed as null in the payload.  
In the graph ql interaction, I can not pass operation name null as it only accepts string ( withOperation(operation: string): this;) while it is defined as  protected operation?: string | null;

Change: Can we allow operation name to accept null?

**Consumer:**
```
export function createTrust(): any {
  return client
    .mutate({
      mutation: gql`
        mutation {
          createTrust {
            id
          }
        }
      `,
      variables: {
        Authorization: MOCK_AUTH_HEADER
      }
    })
    .then((result: any) => result.data)
}
```

**Interaction:**
```
export const createTrustMutaion = new GraphQLInteraction()
  .uponReceiving('a create trust request')
  .withQuery(
    `
    mutation {
      createTrust {
        id
      }
    }
    `
  )
  .withRequest({
    path: '/graphql',
    method: 'POST'
  })
  .withVariables({
    Authorization: MOCK_AUTH_HEADER
  })
  .willRespondWith({
    status: 200,
    headers: {
      'Content-Type': 'application/json'
    },
    body: like(testTrust)
  })
```
Here if I add `withOperation(null)`, typescript compiler complains as the field only accepts string.

### DIFF
--- a/src/dsl/graphql.ts
+++ b/src/dsl/graphql.ts
@@ -25,7 +25,7 @@ export class GraphQLInteraction extends Interaction {
   /**
    * The type of GraphQL operation. Generally not required.
    */
-  public withOperation(operation: string) {
+  public withOperation(operation: string | null) {
     this.operation = operation
 
     return this


mefellows:
Thanks for submitting, I'll see what can be done.

For future reference, please use code gates to style the code - otherwise it becomes unreadable.

mefellows:
@shikhaFNZC apologies for the delay. 

Is there a reason why the operation should be `null` rather than just completely omitted from the payload? I'd prefer just make that an optional call rather than allow a `null` to be passed in, if possible.

mefellows:
@shikhaFNZC bump. Just wanted to confirm the above before I merge this in.

shikhaFNZC:
Hey Matt, Apologies for the delay. Yes Apollo client sends operation name null in request payload if there is no operation name. In the interaction, if there is no operation then pact expects null and the verifications fails.

Although my team has started using operation name from Apollo client so this scenario doesn't arise.

PR #363, opened 2019-09-24
Run contract tests using Jasmine

leonleo997:
The command to run contract tests using Jasmine  `test:jasmine` was running test with mocha instead of jasmine


TimothyJones:
Good spotting, thanks for this.

However, this change currently makes the build fail - you can see the log [here](https://travis-ci.org/pact-foundation/pact-js/jobs/589134042). If you're able to fix this, we can get it merged.

PR #365, opened 2019-09-25
fix(examples): correct karma/jasmine example to use jasmine instead of mocha

leonleo997:


leonleo997:
@TimothyJones It is fixed but I don't know why is this failing
image

TimothyJones:
It's something to do with the way we have configured AppVeyor to kick off PR builds (it starts two, and immediately cancels one, resulting in the failed status). The failed status will eventually be replaced when the second build finishes. In your case, it looks like it will pass:

https://ci.appveyor.com/project/pact-foundation/pact-js/builds/27649016

Apologies for the confusing status report. Fixing it is on the todo list.

Thanks for the updated PR! I'll merge it when the AppVeyor build completes.

PR #373, opened 2019-10-07
fix(matcher): Update isMatcher to correctly check null values

ngbrown:
This is needed for `extractPayload` to work on objects with fields with `null` values.

### DIFF
--- a/src/dsl/matchers.ts
+++ b/src/dsl/matchers.ts
@@ -270,7 +270,7 @@ export interface MatcherResult {
 }
 
 export function isMatcher(x: MatcherResult | any): x is MatcherResult {
-  return (x as MatcherResult).getValue !== undefined
+  return x != null && (x as MatcherResult).getValue !== undefined
 }
 
 // Recurse the object removing any underlying matching guff, returning


TimothyJones:
Thanks for this, @ngbrown. Can you elaborate a bit on why this change is needed, and potentially add a test case for it, please? 

Also, I think you have a `!=` where you meant `!==`.

ngbrown:
The type of `x` is marked to allow `any`, but then is dotted (`.getValue`) after a type cast.  If it is `null` or `undefined` there is an run-time exception.

I have example data and end up needing to add matchers for things like date.  However, sometimes I want to pass the raw data through a test, and so I use `extractPayload` in my code to get the data without the matchers.

`extractPayload` won't process data where the fields are `null` even if I wanted the field in my example data, but just as `null` or `undefined`.

`(x != null)` is the same as `(x !== null || x !== undefined)` and is an allowed exception to the `===` or `!==` rules in eslint.

mefellows:
Sorry I thought I had responded and accepted this. Bringing this in - thanks.

Issue #374, opened 2019-10-08
Dependency on outdated graphql package


ertrzyiks:
This project depends on graphql 0.13.2 which is outdated. The graphql-tag allows 14.0.0, so it should be fine to update.

## Expected behaviour

package.json dependency on
```
 "graphql": "^14.0.0",
```

## Actual behaviour

package.json dependency on
```
 "graphql": "^0.13.2",
```

TimothyJones:
Fixed by #375

PR #375, opened 2019-10-08
Update version of graphql package

ertrzyiks:
Addreses #374


TimothyJones:
Thanks very much for this!

PR #376, opened 2019-10-09
fix(verifier): Allow to use progress formatter during verification

vgrigoruk:
Depends on https://github.com/pact-foundation/pact-node/pull/189


mefellows:
Thanks. I'm going to close this PR as #377 will remove the duplication altogether, as @TimothyJones has fixed the root cause of this issue in Pact Node anyway.

PR #377, opened 2019-10-10
Remove need to duplicate `VerifierOptions`

mefellows:
Removes duplicated VerifierOptions fixed in https://github.com/pact-foundation/pact-node/pull/178.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -167,9 +167,9 @@
       }
     },
     "@pact-foundation/pact-node": {
-      "version": "9.0.3",
-      "resolved": "https://registry.npmjs.org/@pact-foundation/pact-node/-/pact-node-9.0.3.tgz",
-      "integrity": "sha512-RP1+mRb7/Nz/2kqlN22fBchCsurNEYPqJN1GxXiW7cj/I2Uyk72xYUqiD4TUtPWrP8fueR3bvovVDO3QcNKq8g==",
+      "version": "9.0.6",
+      "resolved": "https://registry.npmjs.org/@pact-foundation/pact-node/-/pact-node-9.0.6.tgz",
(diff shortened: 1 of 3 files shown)


TimothyJones:
From your commit message:

> technically a breaking change 'pactBrokerBaseUrl' is a deprecated and useless field. Need to consider what we do here when we release

I think we probably should update all the types in one release and bump the major version - rather than just verifier options. 

I'm not very typescript savvy, but I think removing all the repeated types means anyone using the types will need to make code changes?

It would be nice if we could warn users about the deprecation, but I'm also ok with a release notes that say what the new field is called 

(as an aside, I think there's some inconsistency around the naming of broker fields in pact-node- maybe we could/should clean that up at the same time?)

vgrigoruk:
Could we also bump pact-node to 9.0.6 here, so we get  https://github.com/pact-foundation/pact-node/pull/189 ?

vgrigoruk:
@TimothyJones @mefellows can I help somehow to get this merged?

mefellows:
> I'm not very typescript savvy, but I think removing all the repeated types means anyone using the types will need to make code changes?

Yes, if they were using the old type - it would have to be a major version bump (even though most would use the implicit type required here - as you can see, I didn't need to change any of the regression Typescript test cases but they still work).

> It would be nice if we could warn users about the deprecation, but I'm also ok with a release notes that say what the new field is called

👍

> (as an aside, I think there's some inconsistency around the naming of broker fields in pact-node- maybe we could/should clean that up at the same time?)

There is, but I think we should defer that re-organisation when we move to the new Rust binary because that _will_ likely be a significant change to the API.

In order to keep things moving I think I'll just merge this in for now.

mefellows:
@vgrigoruk I put out a new release this morning, with the upgraded pact-node dependency. Thanks for your patience.

Issue #381, opened 2019-10-18
Pact tests fail when ipv6 is disabled


tristanHessell:
## Software versions

* **OS**: Ubuntu MATE 18.04.2
* **Consumer Pact library**: e.g. Pact JS v9.2.0
* **Provider Pact library**: 
* **Node Version**: v12.4.0

## Expected behaviour

The pact tests to run through as expected.

## Actual behaviour

Pact Consumer tests are failing at the provider setup level with error `listen EADDRNOTAVAIL: address not available ::1:5000`.

For clarity, 5000 is the port i have specified in the `Pact` constructor options.

## Steps to reproduce

Make sure that ipv6 is disabled - for me in linux-land this was through `/etc/sysctl.conf` having the entry `net.ipv6.conf.lo.disable_ipv6 = 1`.

---

The failures were due to my machine not allowing ipv6 on localhost.

These failures were caused by the addition of testing for the ipv6 localhost address (::1), introduced in here: https://github.com/pact-foundation/pact-js/compare/v9.1.1...v9.2.0 in `src/common/net.ts`.

I'm raising this not because there is a bug, but:

1. This has inadvertently "broken" between pact releases.
2. If anyone else has the same error, they'll be able to find this issue.

tristanHessell:
I'm thinking of adding some custom error messaging, similar to how `EADDRINUSE` is handled in https://github.com/pact-foundation/pact-js/blob/3b39a8e0875989963dc420f070bb3a88864a7f54/src/common/net.ts.

Does that sound acceptable?

TimothyJones:
That sounds like a clear improvement, yes please. Thanks for raising this!

mefellows:
My understanding is that you literally can't use pact if IPv6 is disabled and you specify a dedicated port, which is...bad to say the least.

Your suggestion is welcomed and the detailed report muchly appreciated indeed.

I'm left wondering what a good behaviour is in that port check:

1. We could warn if one of the interfaces is not available but others are, and fail if all promises reject (i.e. all checks fail).
1. We could make the interfaces configurable 
1. We could just disable IPv6 entirely

I'm sure there are others too. Thoughts?

tristanHessell:
@mefellows Is there a reason why pact is checking the different hosts as a part of port availability?

Could it not just check the `host` that is specified in the pact constructor (or the default host of `127.0.0.1`)?

[...]

PR #389, opened 2019-10-29
fix(net): stop port check from failing due to ipv6 unavailability (#381)

tristanHessell:
As discussed in issue #381, 

> We could warn if one of the interfaces is not available but others are, and fail if all promises reject (i.e. all checks fail).

---

The code that calls the `close` functions on the server instances used in testing isn't the cleanest approach - more than happy to take on feedback (which also applies to the whole PR).

### DIFF
--- a/src/common/net.spec.ts
+++ b/src/common/net.spec.ts
@@ -41,33 +41,31 @@ describe("Net", () => {
       afterEach(done => closeFn(done))
     })
 
-    // this test uses the test port on all ipv4 and ipv6 hosts via the respective broadcast
-    // addresses (0.0.0.0 and ::)
+    // in all node versions (at time of writing), using the IpV6 broadcast address (::) will
+    // listen on all IpV6 *and* IpV4 addresses (for the specified port) (AKA dual-stacking).
+    // unless `ipv6Only: true` is used in `server.listen({})`. We cannot use this flag as it
+    // has been introduced in node v11.X.
+    // Dual-stacking is configurable at the OS level, but we should not rely on that for the
+    // tests.
     context("when the no local hosts are available", () => {
-      let closeIpV4ServerFn = (cb: any) => cb()
-      let closeIpV6ServerFn = (cb: any) => cb()
+      let closeFn = (cb: any) => cb()
 
-      it("return a rejected promise", () =>
-        createServer(port, defaultHost).then((ipv4Server: { close(): any }) => {
-          closeIpV4ServerFn = ipv4Server.close.bind(ipv4Server)
-
-          return createServer(port, "::").then(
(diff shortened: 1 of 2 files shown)


TimothyJones:
Thanks for this! I agree, there's cleanup possible in the other files too. We shouldn't be calling `done` as well as returning promises (I think returning the promise is probably the best practice).

Also, would you be able to `npm run lint:fix`, please? The build is failing because of the prettier check.

TimothyJones:
Whoops, commented on the wrong PR 🤦‍♂ 

This is much appreciated! Looks like there are some build failures - I'll take a quick look.

src/common/net.spec.ts
```diff
@@ -9,43 +9,88 @@ chai.use(chaiAsPromised)
...
+      let closeIpV6ServerFn = (cb: any) => cb()
+
+      it("return a rejected promise", () =>
+        createServer(port, defaultHost).then((ipv4Server: { close(): any }) => {
+          closeIpV4ServerFn = ipv4Server.close.bind(ipv4Server)
+
+          return createServer(port, "::").then(
+            (ipv6Server: { close(): any }) => {
+              closeIpV6ServerFn = ipv6Server.close.bind(ipv6Server)
+              return expect(isPortAvailable(port, "127.0.0.1")).to.eventually.be
```
TimothyJones: I think this is meant to be `"::"`, not `"127.0.0.1"` ?

tristanHessell: For this test `127.0.0.1` is intended.

---
At line 51, the ipv4 hosts are listened on using the broadcast address (`0.0.0.0`) and line 54 is doing the same for ipv6.

This ensures that all of the checks in `isPortAvailable` will fail.

---
This line (line 57) is just testing with a host that we know is already taken, although I could have used any ipv4 or ipv6 address. However purposely didn't use `localhost`, just to avoid any shenanigans if the person running the test has done something unexpected with their hosts.

TimothyJones: For me (and in CI) this test doesn't pass. Does it pass for you?

tristanHessell: It passes locally for me (on node v12)
I will look into it further - I suspect there is something different between the node versions.

I should be able to explore over the weekend - will keep you posted

tristanHessell: Ooft. The flag `ipv6Only: true` in server.listen({}) only got introduced in node version 11.
My tests and implementation are leaning on that flag being used.

---
Apologies! Node version compatibility totally slipped my mind, I'll get that sorted ASAP

TimothyJones: Much appreciated, thanks!

TimothyJones:
Thanks very much for this!

PR #390, opened 2019-10-29
chore(examples): update Jest Example

yosevukilonzo:
Remove unnecessary done callbacks in async tests since Jest resolves the promises when they are
returned.

### DIFF
--- a/examples/jest/__tests__/index.spec.js
+++ b/examples/jest/__tests__/index.spec.js
@@ -35,7 +35,7 @@ describe("Dog's API", () => {
     })
 
     // add expectations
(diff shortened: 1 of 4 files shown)


TimothyJones:
Thanks for this! I agree, there's cleanup possible in the other files too. We shouldn't be calling `done` as well as returning promises (I think returning the promise is probably the best practice).

Also, would you be able to `npm run lint:fix`, please? The build is failing because of the prettier check.

yosevukilonzo:
Got it. Will plan to update soon and clean up the done calls in other files too!

mefellows:
Nice - thanks!

PR #398, opened 2019-12-12
feat: add email address matcher

aerolalit:


mefellows:
Thanks for the PR!

I'll take a look a little later. I like that you've kept the regex simple, but I also expect people to complain that it doesn't match _their_ expectation of the regex.

mefellows:
So the consensus seems to be going with the regex in this link: https://regular-expressions.mobi/email.html?wlr=1. Would you mind updating the regex with this?

Also, could you please add the new matcher description to the README? Thanks!

mefellows:
Thanks. Bringing this in now, and will update the regex etc. once in.

mefellows:
Release on the way (`9.6.0`) with the email address matcher.

Issue #397, opened 2019-12-08
EventEmitter memory leak warning in jest


ertrzyiks:
## Software versions

* **OS**: Mac OSX 10.14
* **Consumer Pact library**: Pact JS v9.5.0
* **Provider Pact library**: none
* **Node Version**: v12.10.0

## Expected behaviour

No memory leak warning in jest

## Actual behaviour

Spec execution with jest is accompanied with the warning:
```
(node:99515) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 drain listeners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
(node:99515) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
(node:99515) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
```

## Steps to reproduce

https://github.com/ertrzyiks/jest-pact-memory-leak-example

mefellows:
Thanks @ertrzyiks, appreciate the detailed feedback. I'll add this to the backlog for review and update. If you're interested, I'd happily accept a PR for the fix.

ertrzyiks:
@mefellows I could give it a shot, I just need some more time to wrap my head around the codebase. I suppose the fix would be to avoid the side effects of `require`. 

Meanwhile, I find this issue as an extra motivator to go for parallel testing which would be faster and free from the warning.

TimothyJones:
You might also be interested in some of the work @YOU54F and I have been doing with Jest and Pact [here](https://github.com/YOU54F/jest-pact) - this allows parallel tests.

sankalan-sharechat:
Hey @ertrzyiks I would also love to help/tag along to understand how you fix this so that I can also start contributing to pact. Lemme know if I can help :) 
I could also help you document the changes you make.

ertrzyiks:
@sankalan-sharechat I created a work in progress branch https://github.com/pact-foundation/pact-js/pull/399

I realized that removing `.pipe()` call eliminates the memory leak.  Given that last release of https://www.npmjs.com/package/bunyan-prettystream happened 6 years ago before jumping into fixing the issue on its own I consider migrating away from it first. To do that I need to:
 - figure out what is the purpose of this library and what is the behavior we expect from it
[...]

PR #399, opened 2019-12-14
Eliminate memory leak in logger for jest

ertrzyiks:
Related to #397

`.pipe()` call attaches events listeners to the writeable stream
which in case of stdout and sequential jest tests causes memory
leak as those listeners are added for every test file.

Since stdout stream is special, for example it never ends while
the process is runing, we can use simplified piping by
listening to the `data` event of the readable stream and
redirect data to stdout. This way no listeners are leaking.

### DIFF
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -3,7 +3,14 @@ const PrettyStream = require("bunyan-prettystream")
 const pkg = require("./metadata")
 
 const prettyStdOut = new PrettyStream()
-prettyStdOut.pipe(process.stdout)
+
+prettyStdOut.on("data", (data: string) => {
+  process.stdout.write(data)
+})
+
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks @ertrzyiks, this looks good. When I get a moment I'll have a closer look and bring it in if no issues. Much appreciated.

TimothyJones:
Looks good to me! Thanks very much!

TimothyJones:
Ah, I tried this out using `npm link` on an old repository that had an event emitter warning ([this one](https://github.com/TimothyJones/PlusOne)), but the warning persisted. I'm afraid I don't have time to look in to it in more detail at the moment, and it's possible I missed something when I set up the link (or pact - it's from a while ago).

mefellows:
Closing until somebody can pick it back up again.

Issue #433, opened 2020-04-16
RegEx of Matchers.iso8601DateTimeWithMillis is too unflexible


HaVonTe1:
## Software versions

- **Consumer Pact library**: e.g. Pact JS v2.6.0
- **Node Version**: `node --version`
❯ npm -version
6.14.4

    "@pact-foundation/karma-pact": "^2.3.1",
    "@pact-foundation/pact-node": "^10.8.0",
    "@pact-foundation/pact-web": "^9.9.2",

I tried to use the Date-Matcher in the ResponseBody like this:
```
                .willRespondWith({
                      status: 200,
                      headers: {
                        'Content-Type': 'application/json'
                      },
                      body:
                          [{
                            myDate: Matchers.iso8601DateTimeWithMillis('2020-04-08T13:27:34.579Z')
                          }]
                    }
                ))

```
The generated pact.json contains the following regex:
```
          "$.body[0].errorSince": {
            "match": "regex",
            "regex": "^\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\d\\.\\d{3,}([+-][0-2]\\d:[0-5]\\d|Z)$"
          }

```
But the Provider Test written with pact-jvm and Spring Boot created a Date String like this:
2020-04-08T13:27:34.579+0000
And this doesnt match the regex. The problem is the ":" in the offset postfix. 
When I change the regex to the following:
^\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\d\\.\\d{3,}([+-][0-2]\\d(:?[0-5]\\d)?|Z)$
it works like a charm

From what I can see from the ISO8601 dateformat spec this ":" is optional. 

So can we please change the regex here
https://github.com/pact-foundation/pact-js/blob/e0c1b33b3e1da72827a41aceb8cad620623e7719/src/dsl/matchers.ts#L18
To the one above?

mefellows:
PR is welcomed!

PR #434, opened 2020-04-16
#433: changing regex for ISO8601 Dateformat to be more flexible

HaVonTe1:
https://github.com/pact-foundation/pact-js/issues/433


mefellows:
Thanks!

PR #438, opened 2020-04-18
fix: Need to initialise the data otherwise the server is going to load empty

rafaelaazevedo:
I was running these tests and noticed that when running the `npm run test:provider` the server was always empty and not able to create the data, this is because we need to importData() before running the server on the provider spec. Otherwise the server is going to be always empty.


mefellows:
Thanks - yes I guess you're right. It was really just designed for the tests (which do this) but for standalone use, it's pretty useless without the data!

PR #439, opened 2020-04-18
feat: remove need for pact-node in almost all uses, including examples

mefellows:
Creates a new interface that's a thin wrapper over `pact-node`, `Publisher`, that may be used to publish to the broker.

All examples and documentation now updated, to remove references to the project to avoid confusion.


mefellows:
FYI one of the branches failed due to an issue with the travis build setup, which I've fixed in master now. I think. Finally. For good.

Appveyor is failing for the same reason travis was - it's not linking properly (or rather, it's not attempting to do it at all). I'll add an issue to pick that up.

Issue #442, opened 2020-04-23
Update examples to use best practise consumer version and tags


bethesque:

TimothyJones:
Thank you!!! I had this on a mental todo list, but completely forgot about it. It's good to have it tracked.

Also, sometime I want to redo the diagrams on the version page, so that the numbers are semver compliant (they *almost* are).

mefellows:
Merged #443, closing - thanks Beth.

PR #443, opened 2020-04-23
docs: update example publishing scripts

bethesque:
Show best practice for consumer versions and tags.

### DIFF
--- a/examples/serverless/publish.js
+++ b/examples/serverless/publish.js
@@ -1,25 +1,30 @@
-if (process.env.CI !== 'true') {
-  console.log("skipping Pact publish as not on CI...");
+if (process.env.CI !== "true") {
+  console.log("skipping Pact publish as not on CI...")
   process.exit(0)
 }
 
-const pact = require('@pact-foundation/pact-node');
-const childProcess = require('child_process')
+const pact = require("@pact-foundation/pact-node")
+const childProcess = require("child_process")
 
-const exec = command => childProcess.execSync(command).toString().trim();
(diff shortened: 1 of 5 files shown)


bethesque:
https://github.com/pact-foundation/pact-js/issues/442

examples/serverless/publish.js
```diff
@@ -1,16 +1,30 @@
...
+
+//Usually, you would just use the CI env vars, but to allow these examples to run from
+//local development machines, we'll fall back to the git command when the env vars aren't set.
+const gitSha = process.env.TRAVIS_COMMIT || exec("git rev-parse HEAD")
+const branch =
+  process.env.TRAVIS_BRANCH || exec("git rev-parse --abbrev-ref HEAD")
+
 const opts = {
-  pactFilesOrDirs: [path.resolve(__dirname, "pacts/")],
+  pactFilesOrDirs: ["./pacts/"],
```
TimothyJones, line 22: Unless there's a subtlety I am missing, I think the `__dirname` version was more robust, as it is defined to be the same place as this script.

mefellows: That's correct Tim. If somebody runs the script from a different directory it will fail.

mefellows:
Thanks Beth. I've updated to make use of the new Publisher (which now completely hides pact node) and reverted to the uglier, but more resilient, `path.resolve(...)`. People do funny things, like run it from node directly, run npm commands from random directories - it's just easier for us this way!

PR #455, opened 2020-05-13
fix(dsl): Fix extractPayload so that it passes through object properties that are not matchers (fixes #454)

### LINKED ISSUE #454: Matchers.extractPayload ignores properties that aren't matchers
## Software versions

- **Consumer Pact library**: e.g. Pact JS v9.10.0

We have a helper called `extractPayload`, used for creating test objects from pact matchers:

```
    describe('when there are only matchers', () => {
      const rawPlayer = { name: 'tim' };
      const matchedPlayer = { name: Matchers.like('tim') };

      it('Returns the raw object', () => {
        expect(Matchers.extractPayload(matchedPlayer)).toEqual(rawPlayer);
      });
    });
```

However, when there are fields that aren't matchers, they are omitted from the extracted payload:

```
    // These tests fail
    const rawPlayer = { name: 'tim', code: 'THIS_GUY' };
    const matchedPlayer = { name: Matchers.like('tim'), code: 'THIS_GUY' };

    describe('when there are no matchers at all', () => {
      it('Returns the same object', () => {
        expect(Matchers.extractPayload(rawPlayer)).toEqual(rawPlayer);
      });
    });

    describe('when there some fields without matchers', () => {
      it('Returns the raw object', () => {
        expect(Matchers.extractPayload(matchedPlayer)).toEqual(player);
      });
    });
```

## Expected behaviour

Fields that are explicit should be returned without modification (ie, the tests above should pass)

## Actual behaviour

Fields with no matchers are deleted in the returned object (ie, the tests above do not pass)

## Steps to reproduce

Run the tests above

TimothyJones:
Fixes #454 by adding tests and changing the implementation of `extractPayload`.

### DIFF
--- a/src/dsl/matchers.spec.ts
+++ b/src/dsl/matchers.spec.ts
@@ -596,6 +596,54 @@ describe("Matcher", () => {
     })
 
     describe("#extractPayload", () => {
+      describe("when given an object with no matchers", () => {
+        const object = {
+          some: "data",
+          more: "strings",
+          an: ["array"],
+          someObject: {
+            withData: true,
+            withNumber: 1,
+          },
+        }
+
+        it("returns just that object", () => {
+          expect(extractPayload(object)).to.deep.eql(object)
+        })
+      })
+
+      describe("when given an object with some matchers", () => {
+        const someMatchers = {
+          some: somethingLike("data"),
+          more: "strings",
+          an: ["array"],
+          another: eachLike("this"),
+          someObject: {
+            withData: somethingLike(true),
+            withTerm: term({ generate: "this", matcher: "this|that" }),
+            withNumber: 1,
+            withAnotherNumber: somethingLike(2),
+          },
+        }
+
+        const expected = {
+          some: "data",
+          more: "strings",
+          an: ["array"],
+          another: ["this"],
+          someObject: {
+            withData: true,
+            withTerm: "this",
+            withNumber: 1,
+            withAnotherNumber: 2,
+          },
+        }
+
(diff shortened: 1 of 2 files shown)


TimothyJones:
@mefellows  I'm wondering if `extractPayload` is the best name for this. `stripMatchers`?

Also, if we change the name, we have the option of keeping the old behaviour around, if you think people might be relying on it (I can't think of a usecase though, and it was never documented except in the code. I don't know.).

mefellows:
Looks good to me Tim, a nice cleanup also. `stripMatchers` does sound better, perhaps we can alias the function so that both exist? Although, my instinct is to introduce less public APIs where possible.

bethesque:
Don't know if it helps, but the ruby pact implementation term for this is "reify" - to bring into being. It's what gets called in the pact ruby CLI. It would be nice to use consistent terminology across the implementations. One of the things that bugs me is that each implementation ended up with its own names for each function.

TimothyJones:
I'd completely support bringing all the names together- I think that would be a great candidate for a major release (I've been thinking about removing all the deprecated stuff that we currently warn about, could be good to do at the same time).

I'm not sure about `reify` - I'd like the function to be predictable from the name, and I'm not sure I could do that with `reify`. (this is also the same objection to `extractPayload`).

TimothyJones:
I merged this in without any new names - deferring the rename to another time. Although it's not documented, `extractPayload` was used in the message pact example.

bethesque:
`concretify`? 😆

PR #458, opened 2020-05-27
Attempt to improve PactV3 type definitions

audunhalland:
Not 100% sure that everything is covered, but it should be a bit more complete now.


uglyog:
Thanks! That's one item we had on our todo list we can now tick off :-D

mefellows:
Thanks @audunhalland - really appreciating your contributions to Pact over the past year!!

PR #459, opened 2020-05-28
Feat/v3.0.0 experiment with new API withMockServer

audunhalland:
Hi again, some time ago I discussed with you guys on slack the use case of testing several endpoints together. I think the response was along the lines that code should be refactored so each endpoint can be tested in isolation. But let me give you a very real example of a popular programming pattern what would be near impossible to test: React+Apollo. Modern React code has no layers. The API call is performed right from the react component:

```
const MyComponent = () => {
  const response = useQuery('some graphql query', ...);

  return <react elements>;
}
```

There is nothing in the way of performing more than one query:

```
const MyComponent = () => {
  const response1 = useQuery('query 1', ...);
  const response2 = useQuery('query 2', ...);

  return <react elements>;
}
```

Therefore, I think that the current pact v3 API will be a little awkward to use for a lot of developers.

Here I've done an experiment with a new API, which allows (but IMHO does not encourage) running mock servers in parallel for the same test:

```
const provider1 = new PactV3(...);
const provider2 = new PactV3(...);

it('renders correctly', async () => {
  await withMockServer(async (mock1, mock2) => {
    // the test
  }, provider1, provider2);
})
```

I'm imagining this API replacing the `PactV3.executeTest` method. To be honest I wasn't very fond of this object oriented approach. It may appear that the Pact runs the mock server, but in reality that's not the case, it's the other way around: The framework starts a mock server and the mock server has a dependency to the pact. I think this new API better describes the architecture, and leaving the object oriented approach also enables running several mock servers.

This PR as of now is an experiment and is unfinished. I'm looking forward to your comments on the design.

### DIFF
--- a/src/v3/pact.ts
+++ b/src/v3/pact.ts
@@ -63,7 +63,11 @@ export class PactV3 {
     return this
   }
 
-  public withRequestBinaryFile(req: V3Request, contentType: string, file: string) {
+  public withRequestBinaryFile(
+    req: V3Request,
+    contentType: string,
+    file: string
+  ) {
     this.pact.addRequestBinaryFile(req, contentType, file)
     return this
   }
@@ -84,7 +88,11 @@ export class PactV3 {
     return this
   }
 
-  public withResponseBinaryFile(res: V3Response, contentType: string, file: string) {
+  public withResponseBinaryFile(
+    res: V3Response,
+    contentType: string,
+    file: string
+  ) {
     this.pact.addResponseBinaryFile(res, contentType, file)
     return this
   }
@@ -99,7 +107,9 @@ export class PactV3 {
     return this
   }
 
-  public executeTest<T>(testFn: (mockServer: V3MockServer) => Promise<T>): Promise<T> {
+  public executeTest<T>(
+    testFn: (mockServer: V3MockServer) => Promise<T>
+  ): Promise<T> {
     const result = this.pact.executeTest(testFn)
     if (result.testResult) {
       return result.testResult
@@ -142,10 +152,10 @@ export class PactV3 {
           return Promise.reject(new Error(error))
         })
         .finally(() => {
-          this.pact.shutdownMockServer(result)
+          this.pact.shutdownMockServer(result.mockServer)
         })
     } else {
-      this.pact.shutdownMockServer(result)
+      this.pact.shutdownMockServer(result.mockServer)
       return Promise.reject(result.testError)
     }
   }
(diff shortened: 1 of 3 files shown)


src/v3/pact.ts
```diff
@@ -142,11 +152,113 @@ export class PactV3 {
...
+      const newError = new Error(message)
+      // 'Forward' original stack trace:
+      newError.stack = err.stack
+      throw newError
+    }
+
+    throw err
+  } finally {
+    runningServers.forEach(({ pact, mockServer }) => {
+      ;(pact as any).pact.shutdownMockServer(mockServer)
```
audunhalland, line 228: No idea what's going on with this weird opening semicolon. Somehow `prettier` insists on placing it there.

mefellows: It may be the linter or prettier, I'll need to look at the code. It might be a case it's concerned there is ambiguity and has inserted a semi to prevent any compiler mistakes (the compiler can't always correctly infer the correct automatic semicolon insertion).

In this case, it does look wrong - don't worry for now!

I'll also look at this PR tomorrow with fresh eyes.

TimothyJones:
Thanks for this, @audunhalland! I really appreciate the time and thought you've put into this, and it's especially good to get feedback from pact users who are also using GraphQL.

There's a lot to think about here.

In general (outside of your example), I do think the best practice is to separate out the calls. While I think the DSL should be unopinionated about how you choose to write your tests, I think it should also encourage best-practices. One way to do that is to make it hard to write tests that use anti-patterns (a common mistake it to try to put dependencies between tests, like "create user X" followed by "retrieve user X"). Making it easy to create two mock services might encourage poorly written tests in cases where it would be better to have only one mock service. This might in turn lead to frustration with the framework (which I'm keen to avoid).

On the other hand (and relatedly), I would really like to improve the experience for GraphQL users. Although I don't necessarily agree that modern react has no layers, this particular GraphQL pattern is certainly popular, and it would be a clear win if we can support it better.

I'd like to check out this branch and have a play around with a few possibilities - would it be possible for you to provide an example consumer/provider written in this style? Even just a single query component that illustrates the problem would be a big help.

audunhalland:
I now checked some old React+Apollo tests that I wrote some time ago that used the ruby-based mock server. I actually may have misthinked here, because what that API supports is add several interactions to a _single_ mock server, and that's what those tests were actually doing. I now understand that achieving this with two _different_ mock servers isn't going to work, because they would have different endpoint URLs, but you need to configure Apollo to use a single endpoint. That's why having several interactions worked, but having different mock servers would be a dead end.

I see that PactV3 has been designed with a stricter API to avoid some bad practices (single-interaction only?). So maybe I was wrong about this because I didn't think about the single vs. many interactions problem. Thanks for the reply and feedback!

mefellows:
Outside of the React/Apollo discussion, I still believe we need to support for the "multiple interactions" per scenario. Irrespective of our philosophy around design etc. (whichI think we largely agree around), in the end, we're a framework and need to support a broad range of scenarios. In some cases, refactoring just isn't a viable short-term option but adding in Pact tests provides real value.

So, in my view, we absolutely must allow the following cases:

```
1. Given there are multiple providers for a given scenario
2. Given there are multiple calls to a single provider
3. Given there is a single call to a provider
```

TimothyJones:
@audunhalland - I'm still interested in better supporting your use case, so if you do end up having time to provide an example repo with your pattern, it would be very welcome.

PR #464, opened 2020-06-24
chore: trigger an update of docs.pact.io when any .md file changes

bethesque:


bethesque:
Tested an works like a charm.

Issue #249, opened 2018-11-15
Add logLevel to options of pact.verifyPacts()


tiago-dazn:
Problem:
When customProviderHeaders is used, pact always logs  the its content in plain text in logs as warn. Sometimes these headers contain sensitive information like authorisation, which would be best if not logged at all. 
Solution:
Add capability to control logging level of pact.verifyPacts(opts), then we could decide if sensitive information should be logged or not.

bethesque:
Can you help us understand the risk better?

tiago-dazn:
Sure, in my case we need to send the AWS API Key of our REST APIs, which are logged, not a big problem for dev and test environments, but holds us from doing test in production because we can't have sensitive information logged. In other cases, it could

mefellows:
Understand, my initial reaction was similar to Beth. In _most_ use cases, these tests should occur on a development machine/environment where security shouldn't be an issue. There are some edge cases where I think this could be useful, in particular, when verifying against third party APIs or APIs you have less control over than you'd like, but still want to do contract testing (very much an advanced use case).

bethesque:
We should be able to help you with this - but in the meantime, perhaps you could pipe the output through a grep -v to remove the sensitive logs?

YOU54F:
We noted the same in our CI builds, when calling `pact.verifyPacts(opts)`, it would output the credentials to the command line, in the output.

In order to supress this, we call the *.js file containing the verification code, with a shell script

```
#!/bin/bash

set -o pipefail
node server/providerTests/<fileName>.js | grep -v Created 
````

Setting pipefail, means that any errors from the left hand side of the pipe are carried through, rather than just getting the exit code from the grep command.

This means our circleci builds output the verification results in CI, but without leaking the basic auth creds for the broker.

This was important for me in my own projects as mine are public, for my work, less of an issue as everyone needs credentials to read the build logs, but I am always risk adverse so the less I can leak, the better 👍 

See here for a circleci job implementing the above bash script

https://circleci.com/gh/YOU54F/pact-consumer-example-typescript/517?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


output

```
#!/bin/bash
yarn run pact-verify-provider
yarn run v1.12.3
$ ./src/pact/tasks/verify-provider_pact.sh
[2019-03-24T22:27:04.336Z]  INFO: pact@8.0.5/80 on b66534db4b98: Verifying provider
[...]

PR #489, opened 2020-08-25
fix(logger): replace bunyan with pino

### LINKED ISSUE #397: EventEmitter memory leak warning in jest

neezer:
Replaces Bunyan with Pino.

Per [this comment](https://github.com/pact-foundation/pact-js/issues/397#issuecomment-566014868):

> it would also be nice to be able to configure the logging to write to disk at some point too, so the stdout isn't so polluted for those who want it.

Configurable via a new `PACT_LOG_PATH` env variable. Open to alternative suggestions.

> As long as the DSL is backwards compatible (read logLevel doesn't change) I don't see a huge set of requirements for this. A bonus would be a drop in replacement of the interface in the code base without requiring a replace all for the logger.info type statements.

No changes to any implementation files outside of `src/common/logger.ts`.

---

Notes/Questions:
- [x] I didn't know how to go about updating the `package-lock.json` files in `examples/*`. Is this something done after a publish with the new deps?
- [x] My muscle memory put `@types/pino` in `devDependencies` whereas `@types/bunyan` was previously in `dependencies`; is this an issue?
- [x] I didn't see any code references to `logger.time()` or `logger.logLevelName`, so I assumed they were either (a) specific to Bunyan, or (b) dead code. I didn't port either.
- [x] The output of the log messaging is not 100% the same as what Bunyan was producing. I stayed as much as possible with the defaults from [pino-pretty](https://github.com/pinojs/pino-pretty).
- [x] The "change log level after instantiation" logic isn't super-robust; if you provide some bogus log level value—eg., `bananas`—it'll fall back to `info`, which could be confusing. As Pact consumers don't log directly, I didn't see much value in investing more effort here.


neezer:
>> it would also be nice to be able to configure the logging to write to disk at some point too, so the stdout isn't so polluted for those who want it.
> 
> Configurable via a new PACT_LOG_PATH env variable. Open to alternative suggestions.

Noting that whatever this env var—or mechanism—ends up being, we should add it to documentation.

mefellows:
Awesome, thanks for this @neezer! I'll get this looked into this week.

neezer:
@mefellows Cool. 

BTW, I see Appveyor is failing on this:

```
1) Message consumer tests
        receive dog event
           accepts a valid dog:
   Error: the string "The filename, directory name, or volume label syntax is incorrect.\r\n" was thrown, throw an Error :)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

Doesn't look related to my code changes, as it seems Appveyor hasn't been passing for quite sometime. 🤷

neezer:
Apropos this migration: was looking through other open issues related to logging and saw #249. Not 💯 sure on what the API could look like from Pact's perspective, but Pino [supports redactions](https://getpino.io/#/docs/redaction).

TimothyJones:
I really appreciate the proactive approach, thank you so much! I'll take a look through later tonight.

There's a bit of a subtlety in that the ruby binaries (which pact-js is backed by) _also_ take a log parameter as config. This is exposed via the JS API, which might lead to some surprising results if we add an environment variable too. Ideally, we would hide that complexity from the user, but I don't think it's straightforward to do so.

Both logging and configuration are near the top of the list of "things I'd like to improve in pact-js", so I'm excited that there's some momentum on this.

TimothyJones:
@neezer On the build issue - embarrassingly, we haven't been able to fix it for a while because none of the pact-js core team have windows machines we can use to easily look into that problem. I think it's caused by a version bump in pact-node, but it's not easy to determine the cause from the logs.

However, as of last Friday, I now have a windows dev machine on loan. I'll try to look into that tonight, too.

src/common/logger.ts
```diff
@@ -1,35 +1,37 @@
...
-export default new Logger({
-  name: `pact@${pkg.version}`,
-  streams: [
-    {
-      level: (process.env.LOGLEVEL || "info") as bunyan.LogLevel,
-      stream: prettyStdOut,
-      type: "raw",
+let logger = pino(logOpts, destination)
+
+Object.defineProperties(logger, {
```
TimothyJones, line 22: This changes the contract of the `Logger` type from `pino`, since `level` is only a string there, which means calling code might complain about being unable to invoke a function call on a string. I think we might need to export a modified type.

TimothyJones:
Thanks again for this. I took a look last night, and ticked off some of the checklist:

* I don't believe the logger is exported (at least, not on purpose), so I agree that the types are fine in `devDependencies`. 
* I agree, either a or b are true, and that code is safe to remove
* No stress about slight changes to the log format. I don't think the format is part of Pact's exported contract.

Anyway, I checked this out last night, and applied it to a repo that I know shows the problem ([this one](https://github.com/TimothyJones/PlusOne)). I also applied the change to a local copy of `pact-node`, which I can push up soon (or if you would like credit for the commits, @neezer , I'm happy to wait for a PR - either way is fine with me).

(As an aside, both `pact-node` and `pact-js` use bunyan, so we do have to port this over there too (or, perhaps remove the logger from pact-js, and export the one from pact-node?)

Unfortunately, it doesn't fix the problem in that repo (I had the same experience when looking at #399 ). I'm wondering if there's a subtlety in the way the modules work that is causing multiple loggers to be instantiated, but I haven't had the time to dig deeply or cut down my repo to be a minimal example.

However, even if it doesn't fix the memory leak issue, it's valuable to clean up the duplicated logger code and rationalise the way config works, so I'd like to see something like this merged.

src/common/logger.ts
TimothyJones: Possibly we should export our own type that delegates down to `pino` (or whatever), which would nicely separate the concern, and make any further logging library swaps have less impact.

neezer: Alternatively we could take advantage of the fact that the logger is a singleton, and export an entirely separate function to update the log level. A quick search of the repo shows only 4 lines (3 files) where this function would be invoked, so that changeset is small.

That avoids needing to define and export a custom type. Thoughts?

neezer:
@TimothyJones 

> There's a bit of a subtlety in that the ruby binaries (which pact-js is backed by) also take a log parameter as config. This is exposed via the JS API, which might lead to some surprising results if we add an environment variable too. Ideally, we would hide that complexity from the user, but I don't think it's straightforward to do so.

Not sure I follow exactly. Could you rephrase/elaborate?

Aside: I'm not wed to the use of an environment variable. If there's some other configuration mechanism to hook into, happy to adapt to that.

Pointers to where I can find said config would be appreciated.

> - I don't believe the logger is exported (at least, not on purpose), so I agree that the types are fine in devDependencies.
> - I agree, either a or b are true, and that code is safe to remove
> - No stress about slight changes to the log format. I don't think the format is part of Pact's exported contract.

👍 

> I also applied the change to a local copy of pact-node, which I can push up soon (or if you would like credit for the commits, @neezer , I'm happy to wait for a PR - either way is fine with me).

Meh, whichever. So long as fixes get published, I'm happy.

> Unfortunately, it doesn't fix the problem in that repo (I had the same experience when looking at #399 ). I'm wondering if there's a subtlety in the way the modules work that is causing multiple loggers to be instantiated, but I haven't had the time to dig deeply or cut down my repo to be a minimal example.

Bummer. I'll admit I was lazy and assumed that it would since most of the conversation on #399 pertained to the logger. 

I'll continue investigating #399 after this migration gets over the finish line.

Untitled

neezer:
> perhaps remove the logger from pact-js, and export the one from pact-node?

A subpath import could probably suffice; keeps it off the public API for `pact-node`. Eg.,

```ts
import logger from "@pact-foundation/pact-node/src/logger"
```

TimothyJones:
> Could you rephrase/elaborate?

Sure thing- So, pact-js is backed by pact-node, which in turn is backed by the pact-standalone binaries (which are written in ruby). The modules are named pact-node and pact-js for historical reasons - nowadays, `pact-node` really should be called something like `pact-core` - since it's now just a vehicle for the ruby binaries (and not intended to be directly used except for special cases).

At the moment, each of them can write log files: 

Screen Shot 2020-08-26 at 2 57 54 pm

(red is the relevant standalone binary, blue are the JS modules).

The log file written by the standalone binary is currently controlled by the `log` or `logDir` parameter on the relevant constructor (eg `new Pact({ log: "path/to/log.txt" })`), see here https://github.com/pact-foundation/pact-js#api

Additional configuration of logging location might be confusing for users - since the abstraction of the standalone binary will leak. I think it would be reasonable to expect that if you set `PACT_LOG_PATH`, then you wouldn't need to specify `log` in a constructor.

Ideally, we'd set them to append to the same place, but my gut is that it will produce flakey behaviour on windows, which doesn't like the same files being held open by multiple processes.

Screen Shot 2020-08-26 at 3 08 50 pm

Alternatively, we could have one log from the JS modules, and one from the standalone binary, like this:

Screen Shot 2020-08-26 at 3 09 29 pm

However, I can't see an easy way to make the configuration nice and unsurprising in that design.

Hopefully that's a better explanation of the problem, please let me know if there's anything still unclear 👍

TimothyJones:
Re: config - at the moment, we have three ways to configure pact-js:

* A `config` key in package.json. This is only used to override the pact download location (and the mechanism currently has a bug, see https://github.com/pact-foundation/pact-node/issues/234 ). I would also like to rename this.
* Some environment variables. `PACT_BROKER_BASE_URL`, `PACT_BROKER_TOKEN`, `PACT_BROKER_USERNAME`, `PACT_BROKER_PASSWORD`, and `PACT_BROKER_PUBLISH_VERIFICATION_RESULTS`. These are not read by pact-js, but are read by the standalone binaries. 
* Config objects on the pact-js API functions (this is the common way to configure the binaries).

I would really like to add a `pact` key to package.json, which would sharing of pact config within one project.  For bonus points, this could be picked up by the binstubs that `pact-node` exposes.

src/common/logger.ts
TimothyJones: The function approach sounds very reasonable 👍

neezer:
@TimothyJones Gonna go with the approach of making the update to `pact-node` then importing the logger from there in `pact-js`. Got a PR ready to go shortly [now that my local suite is green](https://github.com/pact-foundation/pact-node/pull/237), so gonna put this one back in draft status until those changes get in.

mefellows:
@TimothyJones what's the current status of this PR? Do you need any help?

neezer:
@mefellows Yeah, would love some help. Had to peel off for a bit for an 8k line PR for work and was on vacation last week. I don't have a Windows machine to dev on so my process has been AppVeyor-driven development, which is not efficient at all.

Pino uses [`sonic-boom`](https://github.com/mcollina/sonic-boom) as the underlying library to handle streams, however I tried manually creating a file stream with `sonic-boom` directly and couldn't get the error replicated. FWIW I've never seen an issue with this on macOS/Linux.

neezer:
@mefellows Oh, whoops—I thought we were on the thread for https://github.com/pact-foundation/pact-node/pull/238. IMO that PR is a blocker to the latest proposed changes in this repo.

mefellows:
Thanks @neezer - no dramas, we can track it from either ticket. I'll chat with @TimothyJones to find a time to get through this.

neezer:
@mefellows I fully intend to return to this issue once my 9-to-5 work load lightens up a bit; kinda in crunch atm.

TimothyJones:
Finally got this one in too! Thanks again, @neezer!

Issue #518, opened 2020-10-23
Support webdav REPORT method


HariBhandari07:
## Issue Classification

- [x] Feature request
- [ ] Bug report

## Feature Request
WebDav method `REPORT` is still not supported and trying to use `REPORT` results in an error like `Failed: You must provide a valid HTTP method: GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS.`

TimothyJones:
`9.13.0` is building now with this fix in it.

It will be available in all your favourite npm repositories in about 10-15 minutes.

PR #519, opened 2020-10-23
feat: allow WebDav REPORT request

HariBhandari07:
fixes https://github.com/pact-foundation/pact-js/issues/518


TimothyJones:
Thank you! I'll get this merged and release immediately.

I think we should probably stop validating these all together. It doesn't seem right that Pact-JS has an opinion on which HTTP verbs are allowed. It's not part of the spec.

PR #523, opened 2020-11-08
Fixed test failure on windows.

siad007:
This PR tries to solve the appveyor issue. It is just a guess and therefore marked as draft.


siad007:
And my guess was wrong - sry

mefellows:
👍 

haha thanks, I was just looking at your PR now. When I saw it I thought to myself "Oh, is that all that was!".

Sadly, I think there is an underlying issue with pact-node with how it calls out to the `pact-message` binary.

PR #525, opened 2020-11-23
Feature: Support Pacts for verification

mefellows:
Adds support for the "pacts for verification" endpoint, adding the following types to verifier:

* `consumerVersionTags?: string | string[]`
* `providerVersionTags?: string | string[]`
* `enablePending?: boolean`
* `includeWipPactsSince?: string`

NOTE: I added some commented out fields to the verifier to highlight areas we need to look at for compatibility with existing API, as a sort of "TODO" list.

BREAKING CHANGE: it's a breaking change to the API interface, but have pluralised the properties to be consistent with the current stable interface. Given this is an unstable release, I think it's OK.

### DIFF
--- a/examples/v3/e2e/test/provider.spec.js
+++ b/examples/v3/e2e/test/provider.spec.js
@@ -14,7 +14,7 @@ describe("Pact Verification", () => {
   it("validates the expectations of Matching Service", () => {
     let token = "INVALID TOKEN"
 
-    let opts = {
+    return new VerifierV3({
       provider: "Animal Profile Service V3",
       providerBaseUrl: "http://localhost:8081",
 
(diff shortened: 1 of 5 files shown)


TimothyJones:
The JS / TS parts of this look good to me, although I'm comfortable enough with rust to check that (but I guess I'll eventually need to be, so let me know if you want it double checked).

I like the commented fields you added - I started saying perhaps the comment about the comments should be in the file as well, then realised it would take only a few more seconds to push a commit that did that. So I did.

Should we merge this?

mefellows:
Thanks Tim. Yep, I'll merge now. This was one of the final pieces before we could remove the "you shouldn't use this" memo - because fetching things from the broker this way is in all of our recommended material, we had to add those big caveats.

It's still missing some features (e.g. message pacts), and we can improve things like warns/errors etc., but the RESTful Pact feature is more or less comparable to what is there now.

So now we can focus on the productionisation aspects.

I started putting together a `DEVELOPER.md` file as part of this PR, collecting my thoughts on how I worked locally to make it happen. I'll tidy that up soon and get it published too.

PR #530, opened 2020-12-10
fix: Expect at least one millisecond digit in iso8601DateTimeWithMillis

supersven:
This reflects how it's defined by ISO 8601.

Unfortunately, I can only find third-party references. I got to this, because I had a validation problem between a Typescript consumer and a JVM producer (with this matcher, of course):

```
Expected "2020-12-10T09:01:29.06Z" to match '^d{4}-[01]d-[0-3]dT[0-2]d:[0-5]d:[0-5]d.d{3,}([+-][0-2]d(:?[0-5]d)?|Z)$'
```

### DIFF
--- a/src/dsl/matchers.spec.ts
+++ b/src/dsl/matchers.spec.ts
@@ -583,6 +583,12 @@ describe("Matcher", () => {
           expect(
             iso8601DateTimeWithMillis("2015-08-06T16:53:10.537357Z")
           ).to.be.an("object")
+          expect(
+            iso8601DateTimeWithMillis("2020-12-10T09:01:29.06Z")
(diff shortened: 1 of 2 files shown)


TimothyJones:
Thanks for this, well spotted! I share your (implied) frustration that the standard isn't public. 

It looks to me like you're right - this document which describes a restriction of ISO 8601 allows different numbers of digits: https://www.w3.org/TR/NOTE-datetime

Thanks also for fixing up the commit message! Much appreciated.

I'll merge this in and get it released today.

supersven:
@TimothyJones ,

Cool, Welcome! 👍

PR #541, opened 2020-12-11
UPDATED | Added Popsicle maxBufferSize setting

jpspringall:
Targeted master, as  not sure where to target, so suspect you'll be change that

### DIFF
--- a/src/common/request.ts
+++ b/src/common/request.ts
@@ -42,6 +42,7 @@ export type methods =
 export class Request {
   private readonly transport = Popsicle.createTransport({
     rejectUnauthorized: false, // Need to tell node to ignore bad ssl cert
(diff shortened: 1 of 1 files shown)


TimothyJones:
Thanks for this! Our changelog is generated from the commit messages - would you be able to force push with a commit message like "fix: Increase maxBufferSize to infinity in popsicle requests" or similar?

```
git commit --amend -m "fix(popsicle): Increase maxBufferSize of requests to infinity"
git push -f
```

(We can also fix this with a squash merge if you prefer, but then github doesn't do the attribution properly)

jpspringall:
Done i Think :-|

TimothyJones:
Thanks very much!

Issue #543, opened 2020-12-21
Proposal: Add NestJS Full Working Example


omermorad:
Hi,

I have recently released a new package that connects Pact's great technology to one of the most powerful frameworks for developing server-side applications in NodeJS and TypeScript - NestJS.

The package (with 98% coverage) offers the use of two types of modules, one for the consumer and one for the provider. The modules are actually wrappers of the pact library and use the original interfaces of PactJS.

I'd love for you to take a look at the project on [Github](https://github.com/omermorad/nestjs-pact) or on [NPM's official website](https://npmjs.com/package/nestjs-pact).

These days I am working on a pull-request in which I added a new folder called `nestjs` to the `examples` folder. It contains a fully end-to-end example (similar to the `e2e` original example) of using each of the modules (consumer and provider) in combination with Jest test runner which is the runner that NestJS guys have chosen to in the framework.

Hope to hear from you guys(!) about what you think of the code quality and practices.

Thanks in advance

TimothyJones:
Thanks for this! I’m not really across NestJS or the patterns it uses (I don’t know much about angular), but adaptors to aid adoption are always useful! It’s awesome to see it’s ready for use!

Can you help me understand the value your package adds over using the DSL directly?

You might be interested to take a look at jest pact: https://github.com/pact-foundation/jest-pact - it removes the need for much of the boiler plate. From the looks of the setup in the packages doing, I don’t think you could use it directly, but it might be neat to expose a pactWith with a similar interface for consistency. What do you think?

omermorad:
> Thanks for this! I’m not really across NestJS or the patterns it uses (I don’t know much about angular), but adaptors to aid adoption are always useful! It’s awesome to see it’s ready for use!

Haha thanks a lot! 😎
I must say that after some investigation of Consumer Driven Contracts and Contract Testing in general, `Pact` is actually something we consider to use in the company! 

> Can you help me understand the value your package adds over using the DSL directly?

Yes of course. See, Nest is using a DI system almost identical to the one Angular uses, they use `Injectable` services, or to be more precise, `Injectable` providers.

The two main modules in this package (`Consumer` and `Provider`) wrap `Pact`, `Verifier` and `Publisher`, so you can inject and use some other services/providers while using them.
Let's say for example that I want to create the `Verifier` with some custom options - if I will use the DSL directly, I would not have the ability to inject a `Repository` service or even a `ConfigService` when creating it:

```typescript
@Injectable()
[...]

PR #547, opened 2020-12-25
feat: add nestjs e2e example

omermorad:
Add NestJS fully working, end-to-end example 🎊🎊🎊 

See issue: #543


mefellows:
Hi @omermorad apologies for the delay on this. Xmas break and all of that. I'll aim to get to it this week!

examples/nestjs/consumer/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "prebuild": "rimraf dist",
+    "build": "nest build",
+    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
+    "start": "nest start",
+    "start:dev": "nest start --watch",
+    "start:debug": "nest start --debug --watch",
+    "start:prod": "node dist/main",
+    "test:publish": "node -r ts-node/register test/pact/publish.ts",
+    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
+    "test": "jest && npm run test:publish",
```
TimothyJones, line 18: Ideally, publish would only be run by CI, on distinct versions (or commits) of the source.

examples/nestjs/consumer/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "test:watch": "jest --watch",
+    "test:cov": "jest --coverage",
+    "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
+    "test:e2e": "jest --config ./test/jest-e2e.json"
```
TimothyJones, line 22: I think `test:e2e` isn't meant to be here - the `jest-e2e.json` isn't part of this project. Was it a copy paste error?

examples/nestjs/consumer/src/app.service.ts
```diff
@@ -0,0 +1,74 @@
...
+import { Animal } from './animal.interface';
+
+@Injectable()
+export class AppService {
+  private static readonly AUTH_HEADER = { authorization: 'Bearer token' };
+
+  public constructor(private readonly http: HttpService) {}
+
+  private static getApiEndpoint() {
+    return process.env.API_HOST || 'http://localhost:8081';
```
TimothyJones, line 11: We've spoken about this already, but I think this would be better injected idiomatically, rather than passed through environment variables (apologies if this is done already and I missed the branch or something)

examples/nestjs/consumer/test/app.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+  // Define animal list payload, reusing existing object matcher
+  const animalListExpectation = eachLike(animalBodyExpectation, {
+    min: MIN_ANIMALS,
+  });
+
+  // After each individual test (one or more interactions)
+  // we validate that the correct request came through.
+  // This ensures what we _expect_ from the provider, is actually
+  // what we've asked for (and is what gets captured in the contract)
+  afterEach(() => provider.verify());
```
TimothyJones, line 109: This is specified twice, I think - it's also at the bottom. You only need once (and with `jest-pact`, you don't need it at all).

examples/nestjs/consumer/test/app.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+            },
+            body: animalBodyExpectation,
+          },
+        }),
+      );
+
+      it('returns the animal', async () => {
+        const suggestedMates = await animalsService.getAnimalById(11);
+
+        expect(suggestedMates).toHaveProperty('id', 1);
```
TimothyJones, line 199: You could improve this by using pact's `extractPayload` to strip the example from the matchers:

```
expect(suggestedMates).toEqual(extractPayload(animalBodyExpectation))
```

Also, a little off topic, but I don't think `suggestedMates` is the intended name for this variable - maybe `animal` or `retrievedAnimal`?

examples/nestjs/consumer/test/app.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+
+    describe('and there no animals in the database', () => {
+      beforeAll(() =>
+        provider.addInteraction({
+          state: 'Has no animals',
+          uponReceiving: 'a request for an animal with ID 100',
+          withRequest: {
+            method: HTTPMethod.GET,
+            path: '/animals/100',
+            headers: { Authorization: 'Bearer token' },
```
TimothyJones, line 211: This is a test for a token with the literal content `token`. It might not be what was intended, I'm not sure.

examples/nestjs/consumer/test/app.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+          },
+          body: like(suitor),
+        },
+      }),
+    );
+
+    it('creates a new mate', async () => {
+      return expect(
+        animalsService.createMateForDates(suitor),
+      ).resolves.not.toThrow();
```
TimothyJones, line 255: I think it would be better to check that you unmarshalled the response correctly here - something like `.resolves.toBe(suitor)`

Also, a nitpick- there's a bit of inconsistency in the style of testing promises here - personally, I reckon the tests would read better if all assertions were in one style (also this one is marked `async` but has no await).

examples/nestjs/consumer/test/pact/pact-consumer-config-options.service.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Injectable } from '@nestjs/common';
+import {
+  PactConsumerOptionsFactory,
+  PactConsumerOverallOptions,
+} from 'nestjs-pact';
+
+@Injectable()
+export class PactConsumerConfigOptionsService
+  implements PactConsumerOptionsFactory {
+  createPactConsumerOptions(): PactConsumerOverallOptions {
```
TimothyJones, line 11: As discussed, I think pact's config generally can be improved. 

However, this file includes consumer test config and broker config, which are concepts that aren't needed at the same time, so I think it would be clearer to separate them into their own config files. I think they could be plain objects.

examples/nestjs/consumer/test/pact/pact-consumer-config-options.service.ts
```diff
@@ -0,0 +1,35 @@
+import * as path from 'path';
+import { Injectable } from '@nestjs/common';
+import {
+  PactConsumerOptionsFactory,
+  PactConsumerOverallOptions,
+} from 'nestjs-pact';
+
+@Injectable()
```
TimothyJones, line 8: I don't think this needs to be `Injectable`

examples/nestjs/consumer/test/pact/publish.ts
```diff
@@ -0,0 +1,30 @@
...
+
+(async () => {
+  const app = await NestFactory.createApplicationContext(PactModule);
+
+  const publisher: Publisher = app.get(PactModuleProviders.PactPublisher);
+  const a = 1;
+  // const logger: LoggerService = app.get(Logger);
+
+  // if (process.env.CI !== 'true') {
+  //   console.log('Skipping Pact publish as not on CI');
```
TimothyJones, line 14: This is related to the comment on package json- probably you don't want pact publishing to run when you're not on CI.

examples/nestjs/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "pact-nestjs-full-example",
+  "version": "1.0.0",
+  "scripts": {
+    "test:consumer": "cd ./consumer && npm run test",
+    "test:provider": "cd ./provider && npm run test",
+    "test": "npm run test:consumer && npm run test:provider"
```
TimothyJones, line 7: This actually breaks in CI, because we iterate over all examples and run `npm install`. I reckon we should just have `/examples/nest-js-consumer` and `/examples/nest-js-provider` or similar. What do you think?

examples/nestjs/provider/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "start": "nest start",
+    "start:dev": "nest start --watch",
+    "start:debug": "nest start --debug --watch",
+    "start:prod": "node dist/main",
+    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "test:cov": "jest --coverage",
+    "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
+    "test:e2e": "jest --config ./test/jest-e2e.json"
```
TimothyJones, line 21: As in the other project, I think this script might have been mistakenly included

examples/nestjs/provider/test/pact/pact-producer-config-options.service.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      // Fetch pacts from broker
+      pactBrokerUrl: 'https://test.pact.dius.com.au/',
+      consumerVersionTags: ['prod'],
+      providerVersionTags: ['prod'],
+      enablePending: true,
+      pactBrokerUsername: 'dXfltyFMgNOFZAxr8io9wJ37iUpY42M',
+      pactBrokerPassword: 'O5AIZWxelWbLvqMd8PkAVycBJh2Psyg1',
+      publishVerificationResult: true,
+      providerVersion: '1.0.0',
```
TimothyJones, line 62: This should probably be picked up from the git sha or passed in every time rather than hard coded

examples/nestjs/provider/test/pact/pact-producer-config-options.service.ts
```diff
@@ -0,0 +1,65 @@
+import { Injectable } from '@nestjs/common';
+import { PactProducerOptionsFactory, PactProducerOptions } from 'nestjs-pact';
```
TimothyJones, line 2: Unless I'm missing something, I think these are meant to be named `PactProvider`... not `PactProducer`...  (probably the problem is in nestjs-pact).

TimothyJones (commented):
Nice work! I love the provider side pattern, very slick.

I've been through and made several comments- Not all comments are necessary to address, but I think the driving value should be having examples that can be used as a basis for a new project.

Since most of the comments are quick fixes, I'll take a stab on a branch.

examples/nestjs/consumer/package.json
TimothyJones: Resolved in the other branch

examples/nestjs/consumer/package.json
TimothyJones: Resolved in the other branch

examples/nestjs/consumer/test/app.spec.ts
TimothyJones: Resolved in the other branch

examples/nestjs/consumer/test/app.spec.ts
TimothyJones: I resolved the `extractPayload`, but didn't rename the variable

examples/nestjs/consumer/test/pact/publish.ts
TimothyJones: Resolved in the other branch

examples/nestjs/package.json
TimothyJones: Resolved in the other branch

examples/nestjs/provider/package.json
TimothyJones: Resolved in the other branch

TimothyJones:
Merged in #594

TimothyJones:
Thanks again for this! Much appreciated!!

PR #562, opened 2020-12-29
chore(docs): Correct links to karma examples

individual-it:
the examples live inside the `/examples` folder, so correct the links


TimothyJones:
Thanks for the PR!

PR #563, opened 2020-12-29
[WIP] fix CI

individual-it:


TimothyJones:
Thanks for looking into this! 

The lowest version of node that we support is currently 10, so we definitely should remove the check of node 8.

We used to have two builds, travis (primary) and appveyor (secondary, just a sanity check for windows). Now that travis no longer supports open source builds, we're planning to move to github actions to do primary builds and releases (I plan to do this this week).

So, there's something failing during the message test on windows. However, we haven't had any maintainers with access to windows machines to look into this, and given that most users are not on windows and aren't using message pacts, this hasn't been a high priority (I also now have a windows machine, so hopefully I can take a look soon- once I figure out how to install git and npm, etc).

The failure is fiddly to track down, because the core pact binaries are provided by pact-node, and I think it was a change in pact-node that caused this. Pact-js doesn't pin a specific version of pact-node (for those playing at home, pact-node would really be better named "pact-binaries-ruby", something which I might address this week). At a guess, I would say it might be between pact-node 9.0.2 and 9.0.3, because we changed the way the binaries are spawned in between those versions.

individual-it:
@TimothyJones I have a similar problem, no direct Windows access. I first though it might be just an issue of the old node versions and gave it a go, but as you point it out the issue is more complex

Happy to help out with moving to an other CI, just ping me on slack

I will close this in the meantime

PR #566, opened 2021-01-04
ci: Test and publish with GitHub actions

individual-it:
use github actions to run tests and publish packages to npm & github when a release happened

I could not test the publishing completely because the secrets need to be set


TimothyJones:
This is fantastic, thank you so much!

PR #578, opened 2021-01-18
fix: don't JSON.stringify body if its already a string

### LINKED ISSUE #577: V3 consumer test expects extra `"` for x-www-form-urlencoded bodies
## Issue Classification

- [ ] Feature request
- [x] Bug report

## Bug Report

Having a request in the interaction that specifies `"Content-Type": "application/x-www-form-urlencoded"` and an url-encoded body makes the consumer tests fail with `BodyMismatch`

### Software versions

- **OS**: Ubuntu 20
- **Consumer Pact library**: Pact JS v10.0.0-beta.25
- **Provider Pact library**: Pact JS v10.0.0-beta.25
- **Node Version**: v12.20.1

### Issue Checklist

Confirm the following:

- [x] I have created a reproducable git repository (see below) to illuminate the problem
- [x] I have the read the FAQs in the Readme

### Expected behaviour

tests should not fail when the correct form-data is send

### Actual behaviour

tests fail saying they expect an extra `34` at the beginning and the end of the data

### Steps to reproduce
1. create an interaction with a request that uses content-type and a url-encoded body e.g.
    ```
          body: "first_name=Nanny&last_name=Doe",
          headers: {
            "Content-Type": "application/x-www-form-urlencoded",
          },
    ```
2. create a consumer test that sends those data to the mock server
3. run the consumer tests

#### Example:

1. clone this branch https://github.com/individual-it/pact-js/tree/formDataContenTypeBug (changes made to the `feat/v3.0.0` branch: https://github.com/individual-it/pact-js/commit/0ce2d49d4bd87262b72f39888d4032f0158f4c92)
2. in `examples/v3/e2e` run consumer tests: `npm install; npm run test:consumer`

sending extra quotes `"` at the beginning and the end of the body make the test pass
```
return expect(createMateForDates(
          '"first_name=Nanny&last_name=Doe"',
          () => mockserver.url,
          "application/x-www-form-urlencoded")).to
          .eventually.be.fulfilled
      })
```

### Relevant log files

```
  1) Pact V3
       when a call to the Animal Service is made to create a new mate using form-data body
         creates a new mate with application/x-www-form-urlencoded:
     Error: Test failed for the following reasons:

	Test code failed with an error: expected promise to be fulfilled but it was rejected with 'Error: Internal Server Error'

	Mock server failed with the following mismatches: 
		1) BodyMismatch (at $) Expected text '[34, 102, 105, 114, 115, 116, 95, 110, 97, 109, 101, 61, 78, 97, 110, 110, 121, 38, 108, 97, 115, 116, 95, 110, 97, 109, 101, 61, 68, 111, 101, 34]' but received '[102, 105, 114, 115, 116, 95, 110, 97, 109, 101, 61, 78, 97, 110, 110, 121, 38, 108, 97, 115, 116, 95, 110, 97, 109, 101, 61, 68, 111, 101]'
      at /home/artur/www/pact/pact-js/examples/v3/e2e/node_modules/@pact-foundation/src/v3/pact.ts:152:33
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

individual-it:
if the body is already a string don't try to stringify it

fixes #577 

alternatively we could get rid of `JSON.stringify` all together, the limited e2e tests also pass then

### DIFF
--- a/src/v3/pact.ts
+++ b/src/v3/pact.ts
@@ -136,7 +136,11 @@ export class PactV3 {
   }
 
   public withRequest(req: V3Request): PactV3 {
-    this.pact.addRequest(req, req.body && JSON.stringify(req.body))
+    let body = req.body
+    if (typeof body !== 'string') {
+      body = body && JSON.stringify(body)
+    }
+    this.pact.addRequest(req, body)
     return this
   }
 
--- a/examples/v3/e2e/consumer.js
+++ b/examples/v3/e2e/consumer.js
@@ -60,11 +60,11 @@ const suggestion = (mate, api) => {
 }
 
 // Creates a mate for suggestions
-const createMateForDates = (mate, api = getApiEndpoint) => {
+const createMateForDates = (mate, api = getApiEndpoint, contentType = 'application/json' ) => {
   return request
     .post(`${api()}/animals`)
     .send(mate)
-    .set("Content-Type", "application/json; charset=utf-8")
+    .set("Content-Type", contentType + "; charset=utf-8")
 }
 
 // Suggestions API
--- a/examples/v3/e2e/test/consumer.spec.js
+++ b/examples/v3/e2e/test/consumer.spec.js
@@ -264,6 +264,45 @@ describe("Pact V3", () => {
     })
   })
 
+  describe("when a call to the Animal Service is made to create a new mate using form-data body", () => {
+    const provider = new PactV3({
+      consumer: "Matching Service V3",
+      provider: "Animal Profile Service V3",
+      dir: path.resolve(process.cwd(), "pacts"),
+    })
+
+    before(() =>
+      provider
+        .given("is authenticated")
+        .uponReceiving("a request to create a new mate")
+        .withRequest({
+          method: "POST",
+          path: "/animals",
+          body: "first_name=Nanny&last_name=Doe",
+          headers: {
+            "Content-Type": "application/x-www-form-urlencoded",
+          },
+        })
+        .willRespondWith({
+          status: 200,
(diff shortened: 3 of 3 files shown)


individual-it:
I don't see why CI would fail, but it must be because of the change #579 passes
digging into it

individual-it:
or it is a timeout issue, here is passed in around 8sec
image

so 10 sec timeout might be just to little

mefellows:
Thanks, this looks ok to my eyes.

The failure with the Windows build is a know one (there is an issue with message pact on Windows due to argument spawning). 

I'll look at bringing it in tomorrow.

src/v3/pact.ts
```diff
@@ -136,7 +136,10 @@ export class PactV3 {
   }
 
   public withRequest(req: V3Request): PactV3 {
-    this.pact.addRequest(req, req.body && JSON.stringify(req.body))
+    if (typeof req.body !== 'string') {
+      req.body = req.body && JSON.stringify(req.body)
```
TimothyJones, line 140: Thanks for diagnosing and solving this!

I'd prefer not to modify the request object, as it's a parameter passed in to this function. Would you mind updating this function so it doesn't have side effects on the calling object?

TimothyJones (commented):
Thanks very much for this! One small comment about side effects, other than that this is good to go.

src/v3/pact.ts
individual-it: @TimothyJones good point, I've adjusted the PR

TimothyJones:
Merged! Thanks again!

PR #581, opened 2021-01-18
fix: increase provider test timeout in V3 branch

individual-it:
forward port of #580 to V3 branch


TimothyJones:
Thanks for this! Usually we merge master in to V3, rather than track two fixes. We'll get it in that way.

PR #582, opened 2021-01-19
fix: e2e test examples for V3

individual-it:
1. the local pact file has an other path than given in the comment
2. fix the expected animal-id
    the `.given()` function would set up animal with the id `100` but the test would query and expect id `1` back
    it all worked when running all tests together because id `1` was setup earlier, but the test would fail when running only this test as consumer (`./node_modules/.bin/mocha test/consumer.spec.js --grep 'returns the animal'`) and then verifying the pact by the provider

### DIFF
--- a/examples/v3/e2e/test/provider.spec.js
+++ b/examples/v3/e2e/test/provider.spec.js
@@ -85,7 +85,7 @@ describe("Pact Verification", () => {
       // pactUrls: [
       //   path.resolve(
       //     process.cwd(),
-      //     "./pacts/matching_service-animal_profile_service.json"
+      //     "./pacts/Matching Service V3-Animal Profile Service V3.json"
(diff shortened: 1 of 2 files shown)


examples/v3/e2e/test/provider.spec.js
```diff
@@ -85,7 +85,7 @@ describe("Pact Verification", () => {
       // pactUrls: [
       //   path.resolve(
       //     process.cwd(),
-      //     "./pacts/matching_service-animal_profile_service.json"
+      //     "./pacts/Matching Service V3-Animal Profile Service V3.json"
```
TimothyJones, line 88: Good pickup!

TimothyJones:
This is awesome, thank you! Loving your attention to detail!

PR #584, opened 2021-01-20
fix: Linting code

individual-it:
run `prettier` on the code and enable the lint check in CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "build": "npm run build:v2 && npm run build:v3",
     "build:v2": "npm run lint && npm run compile",
-    "build:v3": "neon build --release",
+    "build:v3": "npm run lint && neon build --release",
     "compile": "rimraf dist && tsc",
     "clean": "rimraf docs dist dist-web coverage .nyc_output logs pacts jscpd.json",
     "clean:native": "rm -rf native/target",
--- a/examples/v3/e2e/consumer.js
+++ b/examples/v3/e2e/consumer.js
@@ -60,7 +60,11 @@ const suggestion = (mate, api) => {
 }
 
 // Creates a mate for suggestions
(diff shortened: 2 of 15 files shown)


TimothyJones:
Hmm. Lint is already run by the build, and lint runs prettier, so if the examples weren't formatted already the build should fail. Probably we have a misconfiguration where the formatter has different settings in checking vs formatting for some reason.

TimothyJones:
Looks like linting has been disabled since march in the V3 branch, in https://github.com/pact-foundation/pact-js/commit/6afc281a80dcc658df6bbb2946a1a122e6014295 . I'm not sure why, but we'll need to re-enable it before merging V3 in.

individual-it:
in this branch it only ran for `build:v2` not `build:v3` https://github.com/pact-foundation/pact-js/pull/584/commits/7d76c63a48288302556d5af4d8cb393b3551231a
and the build script only runs `build:v3` https://github.com/pact-foundation/pact-js/blob/feat/v3.0.0/scripts/ci/build-and-test.sh#L11

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "build": "npm run build:v2 && npm run build:v3",
     "build:v2": "npm run lint && npm run compile",
-    "build:v3": "neon build --release",
+    "build:v3": "npm run lint && neon build --release",
```
TimothyJones, line 10: Hmm. I'm not sure this is the right place for it either, as `neon build` doesn't use anything that would be linted.

I think on master probably the `compile` script should move into the `dist` script (and be removed), and lint is then a `predist` script.

We're really due an overhaul of all these scripts, there are many more than is necessary.

TimothyJones:
> in this branch it only ran for build:v2 not build:v3

Well spotted. However, since the commit I linked, the `lint` script doesn't run lint any more, it only runs the formatter.

PR #585, opened 2021-01-20
chore(ci): run lint before testing

individual-it:
make the promise true: `The recommended approach is to configure Prettier to format on save in your editor ok_hand. If not, don't worry, our lint step will catch it.`
https://github.com/pact-foundation/pact-js/blob/master/CONTRIBUTING.md#code-style-and-formatting

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -19,6 +19,8 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


TimothyJones:
It's already true https://github.com/pact-foundation/pact-js/blob/master/package.json#L8  :) 

The workflow files intentionally call shell scripts so that we can more easily migrate in the future if/when we need to.

individual-it:
OK, I haven't seen that. I noticed v3 branch was not linted. closing this PR and adjusting  #584

PR #590, opened 2021-01-27
fix: don't strigify response that is already a string

individual-it:
### PR Template
analog to https://github.com/pact-foundation/pact-js/pull/578 the response body does not need any stringify if its already a string otherwise an extra pair of quotes is added
```
      AssertionError: expected '"id=100;first_name=Billy;last_name=Goat;animal=goat"' to equal 'id=100;first_name=Billy;last_name=Goat;animal=goat'
      + expected - actual

      -"id=100;first_name=Billy;last_name=Goat;animal=goat"
      +id=100;first_name=Billy;last_name=Goat;animal=goat
```

### DIFF
--- a/examples/v3/e2e/consumer.js
+++ b/examples/v3/e2e/consumer.js
@@ -17,14 +17,20 @@ const availableAnimals = (api = getApiEndpoint) => {
 }
 
 // Find animals by their ID from the Animal Service
-const getAnimalById = (id, api = getApiEndpoint) => {
-  return request
+const getAnimalById = (id, api = getApiEndpoint, format = 'application/json') => {
+  let r = request
     .get(`${api()}/animals/${id}`)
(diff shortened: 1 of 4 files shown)


TimothyJones:
Thanks for this! The change looks good to me, but is there some reason that the example is switching on a query parameter and not the accept header?

individual-it:
I even haven't thought about making the switch by a header, just wanted an easy way to demonstrate the issue
Happy to change it if you feel a header would be better

individual-it:
P.S. I think CI failure is unrelated. I also fails here in a NOP PR https://github.com/pact-foundation/pact-js/pull/591

individual-it:
@TimothyJones I've changed the PR, using `Accept: "text/plain"` now

TimothyJones:
Thanks!

PR #607, opened 2021-02-11
fix: "consumerVersionSelectors" verification option in readme

neildevas:
Fix typo in readme for the provider verification option "consumerVersionSelectors"


TimothyJones:
Good eye! I think this change is right - is there a reason it was closed?

neildevas:
> Good eye! I think this change is right - is there a reason it was closed?

Not a good one - got frustrated that it was failing semantic pull request step :/

TimothyJones:
No problem, we can squash merge when the commits don't follow the convention :+1:

TimothyJones:
Thanks very much for the PR!

PR #610, opened 2021-02-15
fix: upgrade to latest popsicle version to avoid duplicate variable e…

cjr125:
…rror in make-error-cause

### DIFF
--- a/config/webpack.web.config.js
+++ b/config/webpack.web.config.js
@@ -15,12 +15,29 @@ module.exports = {
     filename: "pact-web.js",
   },
   resolve: {
-    extensions: [".ts", ".tsx", ".js"],
+    extensions: [".ts", ".tsx", ".js"], 
+    modules: [
+      path.resolve('./node_modules/popsicle/node_modules'),
+      'node_modules',
+    ],
+    alias: {
+      "form-data": path.resolve('./node_modules/popsicle/node_modules/form-data/'),
+      "tough-cookie": path.resolve('./node_modules/popsicle/node_modules/tough-cookie/'),
+    },
+    fallback: {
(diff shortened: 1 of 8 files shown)


TimothyJones:
Thanks for this! Unfortunately, the tests aren't passing. A quick search suggests that upgrading nock to `11.7.0` would fix this, but unfortunately the tests still failed with a different error after I tried that on this branch.

I also noticed there are several new packages installed in this PR - do we need them? As far as I could tell,  `dgram`,`dns`, `fs`,`http2 `  and `tls` aren't used in this PR.

mefellows:
Thanks for looking Tim (and for submitting @cjr125!). My guess is that those dependencies are there due to TS typing problems or something - but they shouldn't be listed as top level dependencies - perhaps they were just left behind whilst exploring the type definitions that were required?

cjr125:
After submitting this PR, I found a bug with using popsicle and nock together on the newer versions where the error messages don't cooperate. I suspect that I will need to fix the specific version of popsicle that is being used by pact-js and point to it with this PR, but I had some issues getting my npm linking to work. I will update this PR when ready. Thanks for the feedback.

TimothyJones:
Right, that's good to know. I don't think there's any specific reason we're using popsicle (other than we need something that works in the browser and in node), so if it makes sense to replace it, that's fine too.

TimothyJones:
Hi @cjr125 ! Did you have any luck on this one?

cjr125:
@TimothyJones yeah. I got it working with my own custom version of popsicle. I don't think we need to merge it though because the popsicle version is not being maintained any more. I'll close this.

PR #613, opened 2021-02-22
test: fix appveyor build on V3 branch

individual-it:

### PR Template

to fix the appveyor build on V3 branch:
1. don't expect a local brocker
    for linux builds it started in https://github.com/pact-foundation/pact-js/blob/feat/v3.0.0/scripts/ci/build-and-test.sh#L42 but running on windows we cannot start Linux docker images
1. fix env. variables to use correct names from github actions and appveyor not travis

### DIFF
--- a/examples/v3/e2e/test/provider.spec.js
+++ b/examples/v3/e2e/test/provider.spec.js
@@ -9,6 +9,11 @@ server.listen(8081, () => {
   console.log("Animal Profile Service listening on http://localhost:8081")
 })
 
+let pactBroker = "https://test.pact.dius.com.au"
+if (process.env.CI && !process.env.APPVEYOR) {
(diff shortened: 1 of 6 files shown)


TimothyJones:
Cool! This looks good! There are a few uses of `let` that could be `const` - would you mind updating them? That reminds me that we really need to get around to re-enabling lint in this branch.

individual-it:
@TimothyJones I think all the new variables are rewritten in the code, so they should be `let` and not `const` or am I wrong?

TimothyJones:
I think the ones in `publish.js` aren't. Anyway, I don't think it should be a blocker for merging, since we have to do the lint cleanup anyway, and we'll catch any that can be changed then.

Thanks for the PR! Much appreciated!

PR #621, opened 2021-03-03
[V3] Fix query type definition

audunhalland:
It accepts an array when you need multiple query vales that have the same name.


TimothyJones:
Thanks for this! I've also refactored the types a bit to make things much clearer.

PR #623, opened 2021-03-05
fix(package-name): Use the new name (pact-core) for pact-node

TimothyJones:
This PR updates the name for pact-node to the new name of pact-core.

We should not merge it until we have decided on the deprecation strategy for pact-node.


TimothyJones:
Merged this in to `feat/v3`

PR #634, opened 2021-03-23
fix: MatchersV3 don't work for body content

### LINKED ISSUE #632: [V3] pact matchers on namespaced xml elments fail on provider test
### Software versions

- **OS**: ubuntu 20.04
- **Consumer Pact library**: "10.0.0-beta.33"
- **Provider Pact library**:
- **Node Version**: v12.21.0

### Issue Checklist

Please confirm the following:

- [x] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour
Matchers on namespaced xml should work
eg.
``` js
       new XmlBuilder('1.0', '', 'root').build(root => {
          root.setAttributes({"xmlns:h":"http://www.w3.org/TR/html4/"})
          root.appendElement('data', '', data => {
            data.appendElement("h:data", "", MatchersV3.string("random"))
              .appendElement('id', '', MatchersV3.fromProviderState("${id}", "42"))
        })
```
when the server handler is
```
<?xml version="1.0" encoding="UTF-8"?>
  <root xmlns:h="http://www.w3.org/TR/html4/">
      <data>
          <h:data>testData</h:data>
          <id>42</id>
      </data>
  </root>
```
### Actual behaviour

The provider test fail with this error
```
Failures:

1) Verifying a pact between TransactionService and AccountService Given set id - a request to get the plain data returns a response which 
    1.1) has a matching body
           $.root.data.http://www.w3.org/TR/html4/:data.#text -> Expected 'random' to be equal to 'testData'
```

### Steps to reproduce

See the bug reproduced in https://github.com/dpakach/pact-js/pull/2

### LINKED ISSUE #633: [V3] Using fromProviderState on body doesn't work properly
### Software versions

- **OS**: ubuntu 20.04
- **Consumer Pact library**: "10.0.0-beta.33"
- **Provider Pact library**:
- **Node Version**: v12.21.0

### Issue Checklist

Please confirm the following:

- [x] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour
fromProviderState on Body should return proper body
eg.
``` js
 provider
      .given("set id", {id: "42"})
      .uponReceiving("a request to get the plain data")
      .withRequest({
        method: "GET",
        path: MatchersV3.fromProviderState("/data/${id}", "/data/42"),
      })
      .willRespondWith({
        status: 200,
        headers: { "Content-Type": 'text/plain; charset=utf-8' },
        body: MatchersV3.fromProviderState("data: testData, id: ${id}", "data: testData, id: 42")
   })

    return provider.executeTest(async (mockserver) => {
      transactionService.setAccountServiceUrl(mockserver.url)
      return transactionService
        .getText(42)
        .then((result) => {
          console.log(result.data)
          expect(result.data).to.equal("data: testData, id: 42")
        })
    })
  })
```
### Actual behaviour

The body returned is not string, rather a pact object
```
 Expected value   "data: testData, id: 42"
    Received:
      {"expression": "data: testData, id: ${id}", "pact:generator:type": "ProviderState", "pact:matcher:type": "type", "value": "data: testData, id: 42"}
    

```

### Steps to reproduce

See the bug reproduced in https://github.com/dpakach/pact-js/pull/2
Test on https://github.com/dpakach/pact-js/pull/2/files#diff-656f935c93a4d4ea910cb4961f3463c434eb6ea7cbe8d51e2c7b0b3fd4c5ae0cR101

individual-it:
using a matcher in the body does return the whole object but we only want the "example" of it, that is stored in the "value" item.
So when a matcher result is returned then extract the "value" out of it and return that.

Fixes #633

What will happen if your body REALLY contains a body that looks like a Matcher 😱

### DIFF
--- a/src/v3/pact.ts
+++ b/src/v3/pact.ts
@@ -196,6 +196,15 @@ export class PactV3 {
 
   public willRespondWith(res: V3Response): PactV3 {
     let body = res.body
+    if (body !== undefined && body !== null) {
+      if (
+        (body as MatchersV3.Matcher<never>)["pact:matcher:type"] !== undefined
+      ) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        body = body?.value
+      }
+    }
     if (typeof body !== "string") {
       body = body && JSON.stringify(body)
     }
--- a/examples/v3/provider-state-injected/consumer/transaction-service.js
+++ b/examples/v3/provider-state-injected/consumer/transaction-service.js
@@ -33,4 +33,16 @@ module.exports = {
         }
       })
   },
+
+  getXml: (id) => {
+    return axios.get(accountServiceUrl + "/data/xml/" + id).then((data) => {
+      return data
+    })
+  },
+
+  getText: (id) => {
+    return axios.get(accountServiceUrl + "/data/" + id).then((data) => {
+      return data
+    })
+  },
 }
--- a/examples/v3/provider-state-injected/consumer/transaction-service.test.js
+++ b/examples/v3/provider-state-injected/consumer/transaction-service.test.js
@@ -1,12 +1,12 @@
 const path = require("path")
 const transactionService = require("./transaction-service")
-const { PactV3, MatchersV3 } = require("@pact-foundation/pact/v3")
+const { PactV3, MatchersV3, XmlBuilder } = require("@pact-foundation/pact/v3")
 const { expect } = require("chai")
 const { string, integer, url2, regex, datetime, fromProviderState } = MatchersV3
 
 describe("Transaction service - create a new transaction for an account", () => {
   let provider
-  beforeAll(() => {
+  beforeEach(() => {
     provider = new PactV3({
       consumer: "TransactionService",
       provider: "AccountService",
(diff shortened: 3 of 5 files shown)


src/v3/pact.ts
```diff
@@ -196,6 +196,13 @@ export class PactV3 {
 
   public willRespondWith(res: V3Response): PactV3 {
     let body = res.body
+    if (
+      (body as MatchersV3.Matcher<never>)["pact:matcher:type"] !== undefined
+    ) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
TimothyJones, line 203: Is it possible to do this without the eslint-disable/ts-ignore?

TimothyJones:
Are the build failures on this one expected?

individual-it:
@TimothyJones no they are not expected, the hack breaks other tests. I think I need more checks if body is null or undefined

TimothyJones:
I don't think I'm following how the matchers are supposed to work here. @uglyog, can you shed some light?

uglyog:
The problem with this change, while it might solve the problem, drops the matching rules. I think a better solution would be to detect a non-object form (i.e. a primitive value) after processing the body format.

individual-it:
I was wrong about #632, this does not fix it because there the problem is on the provider

individual-it:
I see without the hack the pact file looks like
```
      "response": {
        "body": "{\"pact:matcher:type\":\"type\",\"pact:generator:type\":\"ProviderState\",\"expression\":\"data: testData, id: ${id}\",\"value\":\"data: testData, id: 42\"}",
        "headers": {
          "Content-Type": "text/plain; charset=utf-8"
        },
        "status": 200
      }
    }
```

with the hack:
```
"response": {
        "body": "data: testData, id: 42",
        "headers": {
          "Content-Type": "text/plain; charset=utf-8"
        },
        "status": 200
      }

individual-it:
OK, as I understand it now, this needs to be fixed on the mock-server is that true @uglyog ?

uglyog:
I thought it was an issue in the mock server, but looks like the mock server is behaving correctly, so the problem is not there.

uglyog:
Actually, I think this is due to the content type not being JSON, so it not knowing to process the intermediate format

uglyog:
Closing this as it has been superseded by PR #636

PR #666, opened 2021-05-20
docs: ignore-scripts would break the installation

individual-it:

using `ignore-scripts` option of npm or yarn breaks the installation because the pact standalone binara cannot be downloaded


TimothyJones:
Amazing, thank you!

PR #667, opened 2021-05-24
fix: rename pact-node to pact-core

individual-it:

pact-node was renamed to pact-core (pact-js-core as GH repo) see https://github.com/pact-foundation/pact-js-core/pull/269


TimothyJones:
Ah! Good catch, thank you!

Issue #622, opened 2021-03-05
[V3] allow user to filter by state


mefellows:
### Feature description

The current version of Pact JS allows to run verifications for specific interaction description and/or state: https://github.com/pact-foundation/pact-js#re-run-specific-verification-failures 

A similar approach (env var and possibly a flag in the API) in the V3 branch would be helpful.

The underlying Rust library exposes the following [CLI flags](https://docs.pact.io/implementation_guides/rust/pact_verifier_cli/#command-line-interface):

```
    -c, --filter-consumer <filter-consumer>...
            Consumer name to filter the pacts to be verified (can be repeated)

        --filter-description <filter-description>
            Only validate interactions whose descriptions match this filter

        --filter-state <filter-state>
            Only validate interactions whose provider states match this filter
```

individual-it:
@mefellows something like #668 ?

mefellows:
Closing as 10.x.x supports this.

PR #668, opened 2021-05-24
feat: filter verification by description & state

individual-it:

allow to run verifications for specific interaction description and/or state same as in the current version of Pact JS:  https://github.com/pact-foundation/pact-js#re-run-specific-verification-failures

I confess that I hacked this together without any knowledge of Rust just by staring at the code and googling, so any comment is appreciated
interaction filter if statements taken from https://github.com/pact-foundation/pact-reference/blob/dfd3ba973d22fc092061e8d905d0b4aeba65856b/rust/pact_verifier_ffi/src/verifier.rs#L97-L116 

fixes: #622 

ToDo:
- [x] change docs
- [x] create tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Read [Getting started with Pact] for more information for beginners.
 ```
 npm i -S @pact-foundation/pact@latest
 ```
-Make sure the `ignore-scripts` option is disabled, pact uses npm scripts to download further dependencies. 
+Make sure the `ignore-scripts` option is disabled, pact uses npm scripts to download further dependencies.
 
 ### Do Not Track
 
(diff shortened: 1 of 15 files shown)


mefellows:
Thanks Artur - will take a pass through this tomorrow!

mefellows:
Just noting, I will add these env vars to the Rust code for the verifier.

individual-it:
@mefellows I've adapted the docs and wrote some tests.
How the tests work:
There are some pact files that contain some interactions that would pass and some that would fail if executed. In the test the filter should select only the passing interactions (by setting the env. variable in the spec file via `process.env`. 
So if the test run passes that should prove that the filter works correctly. And if the filter selects the wrong interactions the test run should fail.
Even that is a bit of a weird way to test I choose it (above writing rust unit tests) because as I understood @TimothyJones the rust code would be obsolete soon and that way the tests are implementation agnostic and should prove if the new implementation works correctly or not

individual-it:
Windows tests fail, investigating the reason ...

individual-it:
windows tests pass now after adding a fix in the powershell script that @uglyog already added in the bash script in https://github.com/pact-foundation/pact-js/commit/aac2961e99064b611e1c7d75fe46c5b5f33a3927

mefellows:
Thanks Artur! I'll take a look next week. I'd like to also enable them behind a flag for consistency, so will review it all then.

PR #678, opened 2021-06-02
fix: append text only works with Matchers

### LINKED ISSUE #677: [v3] Using string literals in XmlBuilder gives empty result
### Software versions

- **OS**: ubuntu 20.04
- **Consumer Pact library**:  Pact JS 10.0.0-beta.36
- **Node Version**: v12.21.0

### Issue Checklist

Please confirm the following:

- [x] I have set my log level to debug and attached a log file showing the complete request/response cycle

### Description
Creating interaction with XML body using the XmlBuilder, we cannot use simple strings on the XML
it must always be added as `MaatchersV3.*()`

eg, this interaction
``` 
.willRespondWith({
      status: 200,
      headers: applicationXmlResponseHeaders,
      body: new XmlBuilder('1.0', '', 'ocs').build(ocs => {
        ocs.appendElement('meta', '', (meta) => {
          meta.appendElement('status', '', 'ok')
            .appendElement('statuscode', '', '100')
            .appendElement('message', '', 'OK')
        }).appendElement('data', '', (data) => {
          data.appendElement('id', '', MatchersV3.string(user))
          data.appendElement('display-name', '', MatchersV3.regex(`(${user}|${displayName})`, user))
          data.appendElement('email', '', MatchersV3.regex(`(${user}@example\\..*)?`, ''))
        })
      })
    })
```

### Expected behaviour
```
<?xml version='1.0'?>
  <ocs>
    <meta>
      <status>ok</status>
      <statuscode>100</statuscode>
      <message>OK</message>
    </meta>
    <data>
       <id>test123</id>
       <display-name>test123</display-name>
       <email>test123</email>
     </data>
</ocs>
```
### Actual behaviour
the above interaction will return the body
```
<?xml version='1.0'?>
  <ocs>
    <meta>
      <status></status>
      <statuscode></statuscode>
      <message></message>
    </meta>
    <data>
       <id>test123</id>
       <display-name>test123</display-name>
       <email></email>
     </data>
</ocs>
```
**note that status, status code and message properties are empty**

_fill in here_

### Steps to reproduce

_How can someone else reproduce this bug?_

_For extra bonus points and internet karma, provide a repository, gist or reproducible code snippet so that we can test the problem._

_We recommend forking the project and modifying a relevant example [in the examples folder](https://github.com/pact-foundation/pact-js/blob/master/examples)_

### Relevant log files

_Please ensure you set logging to `DEBUG` and attach any relevant log files here (or link to a gist)._

https://gist.github.com/dpakach/d85379735411a1ee15f1f1308dd14bbd

individual-it:
fixes #677 

there were two problems in the code
1. a typo in the variable name `context` vs `content` so the first `if` wasn't doing anything (introduced by https://github.com/pact-foundation/pact-js/commit/b3b5e6231e9f0ade7be045ff117b441d0169114a#diff-254dbea027a5c57e2b14fb8ff30edc28fea1d39ff2392d95731f7a16f60d5782R70)
2. the second `else if` compiled into the JS: `else if (content)`, so all cases that had any content were send down that road

now the whole section compiles into
```
 if (typeof content === 'object' && content['pact:matcher:type']) {
     this.children.push(new xmlText_1.XmlText(content.value || '', content));
 }
 else {
     this.children.push(new xmlText_1.XmlText(content.toString()));
 }
 return this;
```

I deleted the fist if statement because it wasn't used in the first place and it should not harm to use `toString()` on a string

ToDo:
- [x] add tests

### DIFF
--- a/src/v3/xml/xmlElement.spec.ts
+++ b/src/v3/xml/xmlElement.spec.ts
@@ -1,0 +1,118 @@
+/* eslint-disable @typescript-eslint/dot-notation */
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+
+import * as chai from 'chai';
+import { describe } from 'mocha';
+import { XmlText } from './xmlText';
+import { XmlElement } from './xmlElement';
+import * as MatchersV3 from '../matchers';
+import { Matcher } from '../matchers';
+
+const { expect } = chai;
+
+describe('xml element', () => {
+  describe('appendText', () => {
+    it('can be called with a string', () => {
+      const xml = new XmlElement('my name')
+        .appendText('some string')
+        .appendText('second string');
+      expect(xml, 'XML element').to.have.property('name');
+      expect(xml.name, 'name of XML element').to.equal('my name');
+
+      expect(xml, 'XML element').to.have.property('children');
+      expect(xml['children'], 'children of XML element').to.be.lengthOf(2);
+      expect(
+        xml['children'][0],
+        'type of first child of XML element'
+      ).to.be.instanceOf(XmlText);
+      expect(xml['children'][0], 'first child of XML element').to.have.property(
+        'content'
+      );
+      // @ts-ignore
+      expect(xml['children'][0]['content'], 'content of first child').to.equal(
+        'some string'
+      );
+      expect(xml['children'][0], 'first child of XML element').to.have.property(
+        'matcher'
+      );
+      // @ts-ignore
+      expect(xml['children'][0]['matcher'], 'matcher of the first child').to.be
+        .undefined;
+    });
+
+    it('can be called with a Matcher', () => {
+      const xml = new XmlElement('my name')
+        .appendText(MatchersV3.string('string matcher'))
+        .appendText(MatchersV3.regex(/^.*$/, 'regex matcher'))
+        .appendText(MatchersV3.date('yyyy-MM-dd HH:mm:ss.SSSX'))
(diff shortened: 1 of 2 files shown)


individual-it:
@TimothyJones this is ready for review

maybe the tests check too much private values, but I felt that it should be checked that the xml ends up being what it should.
Alternatively I could stringify the xml and then convert it to JSON, then all the eslint and ts exceptions would not be needed

TimothyJones:
First off, thanks for this! I appreciate all the time you've put in.

> Alternatively I could stringify the xml and then convert it to JSON, then all the eslint and ts exceptions would not be needed

I think that stringification and conversion sounds like a nice approach, as long as it's not creating a lot of code in the test - I'd be concerned about bugs in a complex test harness - who tests the tests? However, as I said in slack, I'm not personally super familiar with patterns for handling xml in JS, so I'm not sure what is easiest.

In terms of review, I found the test a bit hard to follow - maybe it's just a style I'm not used to, but perhaps it would be more self documenting if we could break up the blocks of `expect` into more individual tests with more specific descriptions (even if they're still grouped together).

I'd prefer not to turn off the linter if we can avoid it. If it's necessary to get around the type system for testing, I feel like maybe the type system needs improvement, rather than the tests need to be allowed to break the rules.

That all said, this is an improvement and fixes a bug, so I'm inclined to merge it. If you'd like to rework the tests a bit, that would certainly be welcome. If you end up not being able to get to it, let me know and I'll take a crack.

PR #694, opened 2021-06-15
chore: avoid absolute imports

mduesterhoeft:
Absolute imports can introduce problems for consumers which can be avoided using relative imports. Also relative imports align better with the general style used in pact-js

```
node_modules/@pact-foundation/pact/src/dsl/matchers.d.ts:6:25 - error TS2307: Cannot find module 'common/jsonTypes' or its corresponding type declarations.

6 import { AnyJson } from 'common/jsonTypes';
```

### DIFF
--- a/src/dsl/matchers.ts
+++ b/src/dsl/matchers.ts
@@ -4,8 +4,8 @@
  * but are fixed, to prevent contract invalidation after each run of the consumer test.
  */
 
-import { AnyJson, JsonMap } from 'common/jsonTypes';
 import { isFunction, isNil, isEmpty, isUndefined } from 'lodash';
+import { AnyJson, JsonMap } from '../common/jsonTypes';
 import MatcherError from '../errors/matcherError';
 
 // Note: The following regexes are Ruby formatted,
--- a/src/dsl/message.ts
+++ b/src/dsl/message.ts
@@ -1,4 +1,4 @@
-import { AnyJson } from 'common/jsonTypes';
+import { AnyJson } from '../common/jsonTypes';
 import { Matcher, AnyTemplate } from './matchers';
 
 /**
--- a/src/messageConsumerPact.ts
+++ b/src/messageConsumerPact.ts
@@ -4,7 +4,7 @@
 
 import { isEmpty, cloneDeep } from 'lodash';
 import serviceFactory from '@pact-foundation/pact-core';
-import { AnyJson } from 'common/jsonTypes';
+import { AnyJson } from './common/jsonTypes';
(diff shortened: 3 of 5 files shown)


mefellows:
Thanks @mduesterhoeft, I've just kicked off the build. If it remains green, will bring it in.

TimothyJones:
Could you elaborate on what kind of problems this causes? My understanding is that if absolute imports are causing problems, then there's a mistake in the compiler config (which is quite possibly the case).

I'd actually been planning to clean this up but in the other direction, as relative imports make refactoring annoying in some editor configurations.

TimothyJones:
Just saw your comment in slack with more detail - that definitely shouldn't be happening. I think there's a mistake in tsconfig.

You do raise a good point that the style is inconsistent at the moment. Pragmatically, I'm inclined to merge this because it'll fix your issue and cleanup the style inconsistency. We can then do a housekeeping pass on the tsconfig settings + style later. (relatedly, I'd also like to refactor pact-js so we don't need to import across from other parts of the code, which would make relative vs absolute imports moot)

I really appreciate the active approach of reporting an issue and providing a fix! Thank you!

TimothyJones:
The build is failing because prettier hasn't been run (which enforces formatting consistency). `npm run format:fix` will sort this out for you. I have meetings this morning, otherwise I'd push a commit - if someone else hasn't done it, I'll try to get to it this afternoon.

mefellows:
It's failing on files that aren't relating to this change though, so that is strange. I'll take a look later if I can:


```
[warn] examples/messages/consumer/dog-handler.ts
[warn] examples/nestjs-consumer/test/app.spec.ts
[warn] examples/nestjs-provider/src/app.repository.ts
[warn] examples/nestjs-provider/test/pact/pact-provider-config-options.service.ts
[warn] examples/v3/e2e/test/provider.spec.js
[warn] examples/v3/provider-state-injected/consumer/transaction-service.test.js
[warn] examples/v3/todo-consumer/test/consumer.spec.js
[warn] src/dsl/verifier/proxy/stateHandler/stateHandler.ts
[warn] src/dsl/verifier/proxy/tracer.ts
[warn] src/dsl/verifier/verifier.spec.ts
[warn] src/httpPact.spec.ts
[warn] src/messageProviderPact.spec.ts
[warn] src/v3/verifier.ts
```

mduesterhoeft:
@TimothyJones @mefellows I saw the prettier problems - but because they were unrelated I did not fix them (to not bloat the PR)

Thanks for merging/reacting so fast.

TimothyJones:
This is now out in 10.0.0-beta.38. Sorry it took so long!

PR #695, opened 2021-06-16
fix(verifier): added StateHandler type and improved JSDocs

Lyr-7D1h:
This adds some JSDocs and fixed typing for VerifierV3Options stateHandlers.

### DIFF
--- a/src/v3/verifier.ts
+++ b/src/v3/verifier.ts
@@ -5,6 +5,14 @@ import logger from '../common/logger';
 
 import PactNative from '../../native/index.node';
 
+/**
+ * Define needed state for given pacts
+ */
+export type StateHandler = (
+  setup: boolean,
+  parameters: Record<string, unknown>
+) => void;
+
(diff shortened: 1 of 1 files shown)


TimothyJones:
Thank you! This is much appreciated.

Is `unknown` the right type for the parameters? I had thought that we'd need to add a template for type safety in the parameters, but I haven't had the chance to have a look at it yet.

Please ignore the build failure, it's unrelated (I'm currently working on fixing it)

Lyr-7D1h:
I haven't really used parameters in my pacts so I'm not sure if parameters gets properly validated and what structures it can take. At least `unknown` is type safe and forces users to handle all edge cases. Also forces Typescript 3.0+.

TimothyJones:
Makes sense. This is definitely an improvement over the current typing so I'll merge this in and get it released.

Thanks again for the PR, very much appreciated!

mefellows:
Parameters can accept any valid JSON value

TimothyJones:
This is now released in the latest version.

PR #700, opened 2021-06-23
fix: Avoid exception when printing request.body

lviana-menlosecurity:
When a request is not received we are supposed to get an error message saying "The following request was expected but not received", with the request printed immediately afterwards. However, if the request has a JSON body, we get an exception saying "TypeError: request.body.substr is not a function". The problem is that we need to first convert the JSON body to a string before we can apply the substr method to it (and similarly for the length property). This PR fixes that.

### DIFF
--- a/src/v3/pact.ts
+++ b/src/v3/pact.ts
@@ -102,8 +102,8 @@ function displayRequest(request: MismatchRequest, indent: string): string {
   }
 
   if (request.body) {
+    const body = JSON.stringify(request.body);
     output.push(
-      const body = JSON.stringify(request.body);
       `${indent}Body: ${body.substr(0, 20)}... (${
         body.length
       } length)`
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks! If build passes we can merge.

lviana-menlosecurity:
@mefellows: I caught a typo right after I made the commit, so I made a subsequent commit to fix it. Will a new build be automatically triggered?

mefellows:
> @mefellows: I caught a typo right after I made the commit, so I made a subsequent commit to fix it. Will a new build be automatically triggered?

By default, GH actions don't run for new contributors. I've approved the first, so new PRs will automatically build.

mefellows:
Superceded by #701, thanks for the contribution. Will get another release out today.

PR #702, opened 2021-06-24
Native provider state handlers and request filters

mefellows:
The main changes:

* Replaces the rust callback which doesn't support promises for state handlers whilst preserving the function parameters `setup` and `params`. The response type now expects a Promise.
* Replaces the rust callback which doesn't support promises for request filters
* Updates to logging as I was debugging things
* Updates to types based on current best knowledge (I didn't feel like some of these types belonged in the proxy package, but also, they kind of needed to be there because of import cycles. Feedback welcome here).

Notably, the states (consumer) and state handlers (provider) now use the `JsonMap` type is outgoing/incoming. I think this is correct.

Opening as WIP for feedback, I'll squash it all before merging.

_EDIT_: check is failing due to code coverage. I can improve testing of some of the v3 interfaces, but let's get ✅ on the general approach first.

### DIFF
--- a/examples/v3/e2e/test/provider.spec.js
+++ b/examples/v3/e2e/test/provider.spec.js
@@ -20,49 +20,41 @@ describe('Pact Verification', () => {
     let token = 'INVALID TOKEN';
 
     return new VerifierV3({
+      logLevel: 'info',
       provider: 'Animal Profile Service V3',
       providerBaseUrl: 'http://localhost:8081',
-
-      requestFilter: (req) => {
+      requestFilter: (req, res, next) => {
         console.log(
           'Middleware invoked before provider API - injecting Authorization token'
         );
-
         req.headers['MY_SPECIAL_HEADER'] = 'my special value';
 
         // e.g. ADD Bearer token
         req.headers['authorization'] = `Bearer ${token}`;
-
-        return req;
+        next();
       },
 
       stateHandlers: {
-        'Has no animals': (setup) => {
-          if (setup) {
-            animalRepository.clear();
-            return Promise.resolve({
-              description: `Animals removed to the db`,
-            });
-          }
+        'Has no animals': () => {
+          animalRepository.clear();
+          return Promise.resolve({
+            description: `Animals removed to the db`,
+          });
         },
-        'Has some animals': (setup) => {
-          if (setup) {
-            importData();
-            return Promise.resolve({
-              description: `Animals added to the db`,
-              count: animalRepository.count(),
-            });
-          }
+        'Has some animals': () => {
+          importData();
+          return Promise.resolve({
+            description: `Animals added to the db`,
+            count: animalRepository.count(),
+          });
         },
-        'Has an animal with ID': (setup, parameters) => {
-          if (setup) {
-            importData();
-            animalRepository.first().id = parameters.id;
-            return Promise.resolve({
-              description: `Animal with ID ${parameters.id} added to the db`,
-              id: parameters.id,
-            });
-          }
+        'Has an animal with ID': (parameters) => {
+          importData();
+          animalRepository.first().id = parameters.id;
+          return Promise.resolve({
+            description: `Animal with ID ${parameters.id} added to the db`,
+            id: parameters.id,
+          });
         },
         'is not authenticated': () => {
           token = '';
--- a/src/dsl/verifier/proxy/stateHandler/setupStates.ts
+++ b/src/dsl/verifier/proxy/stateHandler/setupStates.ts
@@ -1,24 +1,58 @@
 import logger from '../../../../common/logger';
-import { ProviderState, ProxyOptions } from '../types';
+import {
+  ProxyOptions,
+  StateFunc,
+  StateFuncWithSetup,
+  ProviderState,
+  StateHandler,
+} from '../types';
+import { JsonMap } from '../../../../common/jsonTypes';
 
-// Lookup the handler based on the description, or get the default handler
+const isStateFuncWithSetup = (
+  fn: StateFuncWithSetup | StateFunc
+): fn is StateFuncWithSetup =>
+  (fn as StateFuncWithSetup).setup !== undefined ||
+  (fn as StateFuncWithSetup).teardown !== undefined;
+
+// Transform a regular state function to one with the setup/teardown functions
+const transformStateFunc = (fn: StateHandler): StateFuncWithSetup =>
+  isStateFuncWithSetup(fn) ? fn : { setup: fn };
+
+// Lookup the handler based on the description
 export const setupStates = (
-  descriptor: ProviderState,
+  state: ProviderState,
   config: ProxyOptions
-): Promise<unknown> => {
-  const promises: Array<Promise<unknown>> = [];
+): Promise<JsonMap | void> => {
+  logger.debug({ state }, 'setting up state');
 
-  if (descriptor.states) {
-    descriptor.states.forEach((state) => {
-      const handler = config.stateHandlers ? config.stateHandlers[state] : null;
+  const handler = config.stateHandlers
+    ? config.stateHandlers[state.state]
+    : null;
 
-      if (handler) {
-        promises.push(handler());
-      } else {
-        logger.warn(`No state handler found for "${state}", ignoring`);
-      }
-    });
+  if (!handler) {
+    if (state.action === 'setup') {
+      logger.warn(`no state handler found for state: "${state.state}"`);
+    }
+    return Promise.resolve();
   }
 
-  return Promise.all(promises);
+  const stateFn = transformStateFunc(handler);
+  switch (state.action) {
+    case 'setup':
+      if (stateFn.setup) {
+        logger.debug(`setting up state '${state.state}'`);
+        return stateFn.setup(state.params);
+      }
+      break;
+    case 'teardown':
+      if (stateFn.teardown) {
+        logger.debug(`tearing down state '${state.state}'`);
+        return stateFn.teardown(state.params);
+      }
+      break;
+    default:
+      logger.debug(`unknown state action '${state.action}' received, ignoring`);
+  }
+
+  return Promise.resolve();
 };
--- a/src/dsl/verifier/proxy/stateHandler/stateHandler.ts
+++ b/src/dsl/verifier/proxy/stateHandler/stateHandler.ts
@@ -1,6 +1,6 @@
 import express from 'express';
 
-import { ProviderState, ProxyOptions } from '../types';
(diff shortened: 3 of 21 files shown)


mefellows:
cc: @lviana-menlosecurity - FYI.

examples/v3/e2e/test/provider.spec.js
```diff
@@ -20,20 +20,18 @@ describe('Pact Verification', () => {
     let token = 'INVALID TOKEN';
 
     return new VerifierV3({
+      logLevel: 'info',
       provider: 'Animal Profile Service V3',
       providerBaseUrl: 'http://localhost:8081',
-
-      requestFilter: (req) => {
+      requestFilter: (req, res, next) => {
```
TimothyJones, line 26: This reminds me that I'm not sure if we are very explicit in the docs about how requestFilter works. I'll open an issue to improve this.

src/dsl/verifier/proxy/stateHandler/setupStates.ts
```diff
@@ -1,14 +1,29 @@
...
 export const setupStates = (
-  descriptor: ProviderState,
+  descriptor: ProviderState | ProviderStateV3,
   config: ProxyOptions
 ): Promise<unknown> => {
   const promises: Array<Promise<unknown>> = [];
 
-  if (descriptor.states) {
+  if (isProviderStateV3(descriptor)) {
+    const handler = config.stateHandlers
```
TimothyJones, line 17: It might be better to do this check a bit more explicitly so that we can warn when no state handlers are present at all (and that check is probably relevant for both V3 and V2).

src/dsl/verifier/proxy/stateHandler/setupStates.ts
```diff
@@ -1,14 +1,29 @@
...
+      ? config.stateHandlers[descriptor.state]
+      : null;
+
+    if (handler) {
+      promises.push(handler(descriptor.action === 'setup', descriptor.params));
+    } else {
+      logger.warn(`No state handler found for "${descriptor.state}", ignoring`);
+    }
+  } else if (descriptor && descriptor?.states) {
     descriptor.states.forEach((state) => {
```
TimothyJones, line 27: I know it was like this when you got here, but I think we should do this as a `map` rather than a `forEach` that produces an array.

TimothyJones: Also, does the V3 verifier sometimes have an array of states? Or does it just fire multiple requests?

src/dsl/verifier/proxy/stateHandler/stateHandler.ts
```diff
@@ -1,6 +1,6 @@
 import express from 'express';
 
-import { ProviderState, ProxyOptions } from '../types';
+import { ProxyOptions, ProviderState } from '../types';
```
TimothyJones, line 3: Is this done automatically by a plugin, and if yes, which one? I'm keen to use it!

src/dsl/verifier/proxy/tracer.ts
```diff
@@ -53,7 +53,7 @@ export const createResponseTracer = (): express.RequestHandler => (
       chunks.push(Buffer.from(chunk));
     }
     const body = Buffer.concat(chunks).toString('utf8');
-    logger.trace('outgoing response', removeEmptyResponseProperties(body, res));
+    logger.trace(removeEmptyResponseProperties(body, res), 'outgoing response');
```
TimothyJones, line 56: Oops, this is meant to be debug. I'll fix that.

src/v3/verifier.ts
```diff
@@ -61,62 +50,124 @@ interface InternalVerifierOptions {
...
+    if (this.config.validateSSL === undefined) {
+      this.config.validateSSL = true;
+    }
+
+    if (this.config.changeOrigin === undefined) {
+      this.config.changeOrigin = false;
+
+      if (!this.isLocalVerification()) {
+        this.config.changeOrigin = true;
+        logger.debug(
```
TimothyJones, line 104: I think this should be `logger.warn`

src/v3/verifier.ts
```diff
@@ -61,62 +50,124 @@ interface InternalVerifierOptions {
...
+
+  // Run the Verification CLI process
+  private runProviderVerification() {
+    return (server: http.Server) =>
+      new Promise((resolve, reject) => {
+        const opts = {
+          providerStatesSetupUrl: `${this.address}:${server.address().port}${
+            this.stateSetupPath
+          }`,
+          ...omit(this.config, 'handlers'),
```
TimothyJones, line 146: You can do this in ES6 without lodash:

```
{ handlers, ...rest } = this.config;
```

TimothyJones (approved):

src/dsl/verifier/proxy/types.ts
```diff
@@ -1,20 +1,50 @@
...
+ */
+export type StateAction = 'setup' | 'teardown';
+
+/**
+ * Define needed state for given pacts
+ */
+export type StateHandlerV3 = (
+  setup?: boolean,
+  parameters?: AnyJson
+) => Promise<JsonMap | void>;
```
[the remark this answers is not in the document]
mefellows: I think they do need to be keyed. I couldn't find definitive proof of this in the specification itself, but all examples I've seen do this.

The Java one has this type definition: https://github.com/pact-foundation/pact-jvm/blob/b301bad9afb2274f7e1edd2df6429ce96e7077de/consumer/src/main/kotlin/au/com/dius/pact/consumer/dsl/PactDslWithState.kt#L55-L58

```kotlin
  fun given(stateDesc: String, params: Map<String, Any?>): PactDslWithState {
    state.add(ProviderState(stateDesc, params))
    return this
  }
```

examples/v3/e2e/test/provider.spec.js
mefellows: Good idea, thanks!

src/dsl/verifier/proxy/stateHandler/setupStates.ts
mefellows: Well, state handlers aren't mandatory - in that you can ignore them and the tests **might** still pass. 

In this way, it warns for each state. Are you suggesting we have a separate one that says "you don't have any state handlers registered, but there are states" type thing?

src/dsl/verifier/proxy/stateHandler/setupStates.ts
```diff
@@ -1,14 +1,29 @@
...
+
+const isProviderStateV3 = (
+  descriptor: ProviderStateV3 | ProviderState
+): descriptor is ProviderStateV3 =>
+  (descriptor as ProviderStateV3).action !== undefined;
 
 // Lookup the handler based on the description, or get the default handler
 export const setupStates = (
-  descriptor: ProviderState,
+  descriptor: ProviderState | ProviderStateV3,
```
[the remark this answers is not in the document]
mefellows: > Edit: I think I like the approach you've taken here the most out of the alternatives I can think of.

Yeah, I actually started with not doing it this way, but it seemed the simplest path forward.

src/dsl/verifier/proxy/stateHandler/setupStates.ts
mefellows: > I know it was like this when you got here, but I think we should do this as a `map` rather than a `forEach` that produces an array.

Just having a look. I think you're probably going for "immutability". Because there are multiple code paths that may not return a promise, we'll need to guard for nulls etc. I think

I'll see what I can do.

> Also, does the V3 verifier sometimes have an array of states? Or does it just fire multiple requests?

It fires multiple, one per state. Technically, the array is only a v3 thing from Ruby, so the Rust verifier will actually never experience this.

Now that I look at the source, I think it will _always_ receive a `ProviderStateV3` so I can actually cleanup this code anyway.

src/dsl/verifier/proxy/tracer.ts
mefellows: I can do it in this PR.

src/v3/verifier.ts
mefellows: That's doing something slightly different (omitting `handlers` altogether), but in this case it actually doesn't matter. I think omit can go entirely, because the rust verifier won't do anything with that prop.

src/dsl/verifier/proxy/types.ts
```diff
@@ -1,20 +1,50 @@
...
+export interface ProviderStateV3 {
+  action: StateAction;
+  params: JsonMap;
+  state: string;
+}
+
+/**
+ * Specifies whether the state handler being setup or shutdown
+ */
+export type StateAction = 'setup' | 'teardown';
```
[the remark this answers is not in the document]
mefellows: I like that, I think it's worth doing now.

src/dsl/verifier/proxy/stateHandler/stateHandler.ts
mefellows: It might be? I have the standard eslint and prettier VS code plugins. 

I also have these relevant workspace settings:

```json
  "typescript.suggest.autoImports": true,
  "typescript.preferences.importModuleSpecifier": "relative"
```

It's also possible during refactoring I removed one of them and brought it back, and it just appended!

src/dsl/verifier/proxy/types.ts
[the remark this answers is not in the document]
mefellows: So I've implemented that suggestion. 

One functional detail I'd like to clarify, is that if you only provide a `StateFunc` (that is, you don't care about setup/teardown) should we invoke it on the setup AND teardown phase? I'm assuming not, and have currently implemented it so that it will _only_ fire on setup. i.e. basically the same behaviour as it does today, except it also optionally takes parameters.

If you want the lifecycle hooks, you should use the variation that has two functions.

Does this seem reasonable to you? It certainly made the examples simpler once I did that, because they no longer needed a guard - so what I'm saying is, nice!

mefellows:
OK I think I've resolved most items @TimothyJones. Feedback on the `StateFunc` is the only remaining thing I think.

I'll then tidy up commits, sync up with `feat/v3.0.0` and merge after that.

src/dsl/verifier/proxy/stateHandler/setupStates.ts
TimothyJones: Yeah, that’s what I was thinking. It feels like a “you probably are missing some config” situation.

mefellows:
Going to merge this in. Now that it's here, adding support for messages is actually not a huge amount of work either.

TimothyJones:
Awesome! Excited about this

PR #704, opened 2021-06-28
feat: reset consumer test state to enable re-use of PactV3 class

mefellows:
Wipes the internal state between test runs to enable re-use of the provider.

Not thread safe (it wasn't before), but if you want to run in parallel you can still create multiple providers as per what is currently the case, and the pact will be merged.

### DIFF
--- a/examples/nestjs-consumer/test/app.spec.ts
+++ b/examples/nestjs-consumer/test/app.spec.ts
@@ -21,8 +21,13 @@ pactWith(
     });
 
     // Alias flexible matchers for simplicity
-    const { eachLike, like, term, iso8601DateTimeWithMillis, extractPayload } =
-      Matchers;
+    const {
+      eachLike,
+      like,
+      term,
+      iso8601DateTimeWithMillis,
+      extractPayload,
+    } = Matchers;
 
     // Animal we want to match :)
     const suitor: Animal = {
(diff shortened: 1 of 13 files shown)


TimothyJones:
Nice! Is configure the best name? It sounds a bit similar to setup to me.

I think the non thread safeness is probably ok because node is single threaded- even when just runs more than one test at once, it’s not sharing memory between the tests. Am I missing anything?

mefellows:
`setup` is probably better. 

> I think the non thread safeness is probably ok because node is single threaded- even when just runs more than one test at once, it’s not sharing memory between the tests. Am I missing anything?

Not really, the main point is that if you want to allow parallel tests, you should create multiple providers. I think that's fixable though, we'd just need to track multiple `pact` instances to specific `executeTest` functions. Separate problem for another day.

Also prettier just lost it with me. I'll update this first, and do a prettier fix in a separate push.

TimothyJones:
I think the prettier problem is something to do with my local setup. CI and your prettier seem to agree, I don't know what's going on locally for me.

mefellows:
no, I re-ran `npm run format:fix` and it "fixed" a bunch of files. Same problem it seems!

TimothyJones:
It was broken after I pushed formatting changes that my local wanted (I had incorrectly assumed someone merged something that didn’t build)

mefellows:
ah, thanks for clarifying!

PR #708, opened 2021-07-08
docs: update readme and examples to use providerVersionTags and consumerVersionTags instead of tag (singular)

bethesque:
Also updates the tag values to use best practice examples.


TimothyJones:
Great catch! Thank you!!

PR #712, opened 2021-07-19
fix: using relative import for provider, fixes problems with intellisense in vscode

Lyr-7D1h:
Fixes vscode intellisense by using a relative import instead of an absolute one.

### DIFF
--- a/examples/nestjs-provider/test/pact/pact-provider-config-options.service.ts
+++ b/examples/nestjs-provider/test/pact/pact-provider-config-options.service.ts
@@ -28,24 +28,24 @@ export class PactProviderConfigOptionsService
           this.animalRepository.clear();
           token = '1234';
 
-          return 'Animals removed to the db';
+          return { description: 'Animals removed to the db' };
         },
         'Has some animals': async () => {
           token = '1234';
           this.animalRepository.importData();
 
-          return 'Animals added to the db';
+          return { description: 'Animals added to the db' };
         },
         'Has an animal with ID 1': async () => {
           token = '1234';
           this.animalRepository.importData();
 
-          return 'Animals added to the db';
+          return { description: 'Animals added to the db' };
         },
         'is not authenticated': async () => {
           token = '';
 
-          return 'Invalid bearer token generated';
+          return { description: 'Invalid bearer token generated' };
         },
       },
 
--- a/src/v3/verifier.ts
+++ b/src/v3/verifier.ts
@@ -1,9 +1,9 @@
 import { isEmpty } from 'ramda';
-import { ProxyOptions, StateHandlers } from 'dsl/verifier/proxy/types';
 
 import * as express from 'express';
 import * as http from 'http';
 import * as url from 'url';
+import { ProxyOptions, StateHandlers } from '../dsl/verifier/proxy/types';
 import { localAddresses } from '../common/net';
 import { createProxy, waitForServerReady } from '../dsl/verifier/proxy';
 


examples/nestjs-provider/test/pact/pact-provider-config-options.service.ts
```diff
@@ -28,24 +28,24 @@ export class PactProviderConfigOptionsService
           this.animalRepository.clear();
           token = '1234';
 
-          return 'Animals removed to the db';
+          return { description: 'Animals removed to the db' };
```
mefellows, line 31: what is the purpose of this change?

mefellows:
@TimothyJones is our stance still OK with sticking with relative imports? I recall having a chat but can't see where.

TimothyJones:
Yep! I think the absolute imports are currently a mistake on my part (although my vscode handles them fine)

examples/nestjs-provider/test/pact/pact-provider-config-options.service.ts
Lyr-7D1h: Tests were failing due to the need of having JsonMap as a return type. Also showing that imports didn't resolve correctly.

Lyr-7D1h:
@TimothyJones for me in the `verifier.ts` it does resolve correctly but when transpiled and used as a library it does not and assigns an any type.

examples/nestjs-provider/test/pact/pact-provider-config-options.service.ts
TimothyJones: This shouldn't break with the change, though - because nestjs-pact only uses the v2 options type, which hasn't changed.

TimothyJones: Really these tests shouldn't be returning anything at all, though - because I don't think it's valid in V2. @mefellows, thoughts?

TimothyJones:
Thanks for the fix! We'll merge it as soon as we've worked out what's going on with that provider test (I think that test is disabled anyway, because it's not compatible with the next release, see https://github.com/omermorad/nestjs-pact/issues/7 ).

As an aside, do you know how to get a typescript module to flatten the types? At the moment, intellisense is happy to deep link into `@pact-foundation/pact/some/path`, which makes it hard to refactor things without it being a breaking change. I'd like intellisense to still be able to reason about the types, but to get them all from the base import.

Lyr-7D1h:
@TimothyJones Add `@internal` to the exports you want to hide. add `stripInternal: true` to tsconfig.json.
image
image
image

It should only use the exports from index.js

From:
https://stackoverflow.com/questions/59122428/typescript-library-hide-internal-exports.

examples/nestjs-provider/test/pact/pact-provider-config-options.service.ts
mefellows: Correct. I actually think it's completely misleading, so would prefer to remove altogether.

TimothyJones:
You're amazing! Thank you for doing the legwork!

TimothyJones:
Aha! I finally had a moment to look at this - the problem was indeed a bug that this fix exposed. 

```
export type VerifierOptions = VerifierV3Options & ProxyOptions; 
```
^ `VerifierOptions` was clobbering the other `VerifierOptions`. I'll rename the types accordingly and get this merged in.

TimothyJones:
Thanks again for the PR! Apologies for taking so long on this one. We'll release it this weekend.

PR #732, opened 2021-08-24
feat: use ffi-core for pact verification

mefellows:
Integrates the new `pact-core` package with the FFI bindings, to help with replacement of neon and Rust code in Pact JS.

Specifically, can be used as a drop-in replacement of the Ruby verification code used currently.

Done:
* Replace the v3 verifier with the FFI bindings
* Add BREAKING CHANGE to commits
* Add a MIGRATION.md to help users migrate between beta version
* Update documentation to show how to use latest version

TODO:
* Fix windows build


TimothyJones:
I think we should also (during beta) pin the version of pact-core. This means it's easier to predict the behaviour combination and means that it's safer to release new point releases while we might make mistakes with the FFI.

TimothyJones:
I'm also curious why the windows build is failing with a DLL error, but we released pact-core without build failures. Maybe windows builds aren't running there any more :/ I'll go look.

Also, why is semantic PR unhappy?

TimothyJones:
@mefellows : In updating the docs, I've noticed that this version doesn't support `disableSSLVerification`. I think we need to add that in pact-core before releasing this. Looks like there's `--disable-ssl-verification` we can use.  

Also, I think the option should be `disableSslVerification`, for consistency with the rest of the camelcased options. I might rename it at the same time, since this is a breaking change anyway.

mefellows:
🎉 Awesome Tim!!

PR #748, opened 2021-09-29
refactor: remove rust code from v3 matchers

mefellows:
V3 matchers will be used between both the existing and new v3 compatible interface. Removing rust is a dependency to migrate to the FFI.

In doing this, the rust generators were removed that created examples based on a regex. The following matchers now must have an explicit example provided:

* `datetime`
* `timestamp`
* `time`

Thank you for making a pull request!


mefellows:
Currently blocked on the ref-napi code.

PR #753, opened 2021-10-01
chore: Remove publish.js in all examples

CalCorbin:
This change removes the references to `publish.js` in the example `package.json` files. Instead, we replace them with examples that use the pact-broker CLI.


CalCorbin:
I have a few questions-

1. Would you like me to replace the `pact/pacts` with a descriptive env variable like `YOUR_PACT_FILES_OR_DIR`?
2. The same goes for the cli command `--broker-base-url`. I could add `BROKER_BASE_URL` or something similar.
3. Do you also want me to remove the `publish.js` files all together from the example directory?

TimothyJones:
1. That's an interesting idea. I'm not sure how we'd do it differently for the different examples (not all examples put the pact files in the same location)
2. I'm not sure whether it's better to put the broker URL in an environment variable or not. I'd happily accept whatever you think is appropriate.
3. Yes please!

We'll also need to include the broker password and username in the environment variables (`PACT_BROKER_USERNAME` and `PACT_BROKER_PASSWORD`). We can set those in the script `scripts/ci/build-and-test.sh` before the examples are run.

mefellows:
I think you'll need to include the credentials in that call also (i.e. https://github.com/pact-foundation/pact-js/blob/master/examples/e2e/test/publish.js#L27)

> Would you like me to replace the pact/pacts with a descriptive env variable like YOUR_PACT_FILES_OR_DIR?

I think it's fine as you have it.

> The same goes for the cli command --broker-base-url. I could add BROKER_BASE_URL or something similar.

Explicit is fine for me.

> Do you also want me to remove the publish.js files all together from the example directory?

Yes please!

mefellows:
ha, Tim was on it too - at least we were mostly consistent :P

CalCorbin:
Thanks all! Question-

I found a reference to a typescript version of `publish.ts` in this example file: https://github.com/CalCorbin/pact-js/blob/a4802313b1d476b70773d62feadf7e47a034458b/examples/nestjs-consumer/package.json#L15

Do we want to replace that command with the same npm script used in the other examples?

CalCorbin:
Alrighty!
1. Removed `publish.js` files.
2. Updated package.json scripts that reference the old publish script, so hopefully CI tests pass now.

CalCorbin:
> We'll also need to include the broker password and username in the environment variables (`PACT_BROKER_USERNAME` and `PACT_BROKER_PASSWORD`). We can set those in the script `scripts/ci/build-and-test.sh` before the examples are run.

My bad, it just dawned on me that I need to make this change. Doing that now.

scripts/ci/build-and-test.sh
```diff
@@ -2,6 +2,9 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+require_env_var PACT_BROKER_USERNAME
```
CalCorbin, line 5: Let me know if this works for env vars on this file for ci.

TimothyJones:
So, you can actually run the `scripts/ci` scripts locally - the design is that the `scripts/*` files are entry points for local developers, and the `scripts/ci` are entry points for the CI. Running locally will reduce the feedback loop on knowing whether or not it passed.

(Aside: I did add instructions that say this, but it looks like they're in the branch for the `v10` release, which confusingly is named `feat/v3.0.0`, because it brings support for pact specification v3. We can take care of the merge into that branch once this is merged)

It's failing at the moment because the script doesn't have the `require_env_var` macro at that point. You can fix this with:

```
. "$DIR"/lib/robust-bash.sh
```

after the `DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"` line

TimothyJones:
Otherwise this is great! I think some of the examples put their pacts in `./pacts`, and some in `./pact/pacts`. I *think* the differentiator will be:

* jest-pact uses `./pact/pacts` by default
* pact otherwise uses `./pacts`
* Some examples may have explicit config to a different path

I think build will fail if any of them are in the wrong place. I like your suggestion of using an environment variable to make it clear that this is a path - one improvement could be to configure them all to `./pact/pacts`, and then set that environment variable in the workflow script too.

Thanks so so much for doing this! It's one of those fiddly tasks that has been on the todo list for ages (even though it was only recently filed), but never makes it to the top of the priority list when I have time to look at pact stuff.

TimothyJones:
>  I found a reference to a typescript version of publish.ts in this example file: https://github.com/CalCorbin/pact-js/blob/a4802313b1d476b70773d62feadf7e47a034458b/examples/nestjs-consumer/package.json#L15
> Do we want to replace that command with the same npm script used in the other examples?

Yes please, that would be awesome.

CalCorbin:
Awesome, thanks for the feedback @TimothyJones! I'll get these changes up some time tomorrow morning in American Central time.

scripts/ci/build-and-test.sh
TimothyJones: Ah, also - you'll need to explicitly set these variables *after* `npm run dist`. The pact cores (ruby or rust) that pact-js uses configure themselves with `PACT_BROKER_PASSWORD` and `PACT_BROKER_USERNAME` - which means that if the variables are set before `npm run dist` then some of the tests may fail.

TimothyJones: `require_env_var` checks to see if a variable is set, but we'll want to actually set them with `export PACT_BROKER_PASSWORD=<whatever it is>`.

I forget the appropriate values, but they're hardcoded in the existing publish.js scripts. Even though they are credentials they are "safe" to be committed- they are for the public test account that's very kindly provided by Pactflow.

TimothyJones:
No problem! Thanks so much for the PR, and have an excellent evening.

scripts/ci/build-and-test.sh
CalCorbin: > I forget the appropriate values, but they're hardcoded in the existing publish.js scripts. Even though they are credentials they are "safe" to be committed- they are for the public test account that's very kindly provided by Pactflow.
 
 This makes sense now, I was really hesitant to commit hard coded values. 😅 
 
 Thank you, I'll get these changes up today!

CalCorbin:
The following changes have been made:

1. Updated the `pact` directory on each example script in `pact:publish`. Ran all of these tests locally and ensured they are passing.
2. Removed the additional `publish.ts` file.
3. Added pact user creds to the `build-and-test` file, tests all passed locally on node 12.

As far moving  `./pact/pacts` to an env var, I'm not as convinced now that it will be useful. Going through and updating the npm scripts made me think perhaps that it would be overkill.

bethesque:
> As far moving ./pact/pacts to an env var, I'm not as convinced now that it will be useful

Agreed.

CalCorbin:
So it looks like the tests are passing, and I'm able to get them all to pass locally. I think Github workflow/CI isn't happy with the `--tag-with-git-branch` command, but taking a guess there. 

This is the error we get with all the tests after they appear to pass in CI:
```
> pact-broker publish ./pacts --consumer-app-version="$(npx @pact-foundation/absolute-version)" --tag-with-git-branch --broker-base-url=https://test.pact.dius.com.au

npx: installed 46 in 2.223s
/home/runner/work/pact-js/pact-js/examples/e2e/node_modules/@pact-foundation/pact-node/standalone/linux-x64-1.88.63/pact/lib/ruby/lib/ruby/gems/2.2.0/gems/bundler-1.9.9/lib/bundler/shared_helpers.rb:78: warning: Insecure world writable dir /opt/hostedtoolcache/node/14.18.0/x64/lib/node_modules/npm/node_modules/npm-lifecycle/node-gyp-bin in PATH, mode 040777
PactBroker::Client::Error - Command `git rev-parse --abbrev-ref HEAD` didn't return anything that could be identified as the current branch.
```


I think the key message is 
```
PactBroker::Client::Error - Command `git rev-parse --abbrev-ref HEAD` didn't return anything that could be identified as the current branch.
``` 
I'll research this today and see if I can come up with a fix. Feel free to chime in if anyone is familiar with this!

CalCorbin:
Alright, I'm fairly certain that the github workflow doesn't know what branch it's on in the CI.

TimothyJones:
Yes, I agree. I think this is something to do with the PR settings in our workflow script- the branches on the repo do know what branch they are. I think it will probably just work if we merge it

I’ll take a look at it tomorrow.

CalCorbin:
Ok, I wrote a potential fix for the workflow test failures in the `build-and-test.yml` file. You can find those changes here: https://github.com/pact-foundation/pact-js/blob/0727843a5732bf6c4811fd3ec26747db868084c8/.github/workflows/build-and-test.yml

Let me know what you all think and if it's worth seeing if that fixes the issue.

.github/workflows/build-and-test.yml
```diff
@@ -16,6 +16,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+          repository: pact-foundation/pact-js.git
+          token: ${{ secrets.GITHUB_TOKEN }}
```
CalCorbin, line 20: I believe that the workflow checks were unable to find the git branch because these parameters were missing in the yml file, but I'm not certain until we run the tests again.

bethesque:
> Alright, I'm fairly certain that the github workflow doesn't know what branch it's on in the CI.

Correct. What we need to do is update the --tag-with-git-branch code to pick up the Github Action env var.

It's this code here https://github.com/pact-foundation/pact_broker-client/blob/master/lib/pact_broker/client/git.rb#L28

bethesque:
In the meantime, this is how I process up the GITHUB env vars to use in the pact code. It cuts off the "/refs/heads" at the start.

```
      - name: Verify pacts
        run: "GIT_SHA=${GITHUB_SHA} GIT_BRANCH=${GITHUB_REF##*/} bundle exec rake pact:verify"
```

https://github.com/pact-foundation/pact_broker/blob/master/.github/workflows/test.yml#L82

CalCorbin:
> 
> 
> > Alright, I'm fairly certain that the github workflow doesn't know what branch it's on in the CI.
> 
> Correct. What we need to do is update the --tag-with-git-branch code to pick up the Github Action env var.
> 
> It's this code here https://github.com/pact-foundation/pact_broker-client/blob/master/lib/pact_broker/client/git.rb#L28

@bethesque I'm looking at the latest commit on `pact_broker-client`, looks like you updated `--tag-with-git-branch` to accept the GITHUB_REF env var now? Hopefully that means this works now. 

I'm going to remove my changes to ` .github/workflows/build-and-test.yml ` in the meantime.

bethesque:
Man, you're on to it quickly :P If it's using the built in standalone, that will need to be updated in the pact-core library first. If it's using the docker image, it's available now.

CalCorbin:
Looks like its using the pact-core library. No idea how to update that in the examples unfortunately.

bethesque:
Looks like pact-core will need to be released first, before the --tag-with-git-branch will work. I think that's a bit of a manual process at the moment, so I'd recommend using the `GIT_BRANCH=${GITHUB_REF##*/}` trick until that's released, so we can merge this.

TimothyJones:
Wow, thanks for looking in to this! I think the problem is that the workflow that's used is the one in the master branch, not the one here - I'll check that now.

TimothyJones:
@bethesque : I'm confused because pact-js *does* check out the repository correctly. `git describe` is what backs the versioning tool that I use in the examples.

TimothyJones:
Anyway, this is excellent, and will close the ticket. I'll deal with any build problems (which are kind of unrelated) after the merge.

Thank you so much!

CalCorbin:
Awesome, thank you all so much. I enjoyed it!

PR #784, opened 2021-12-30
Fix interaction response status type

beckerei:
Currently, it is allowed to write

```ts
willRespondWith: {
  status: Matchers.like(200)
},
```

but the result will always be 500 as Matchers are not supported for the response status.

### DIFF
--- a/src/dsl/interaction.ts
+++ b/src/dsl/interaction.ts
@@ -22,7 +22,7 @@ export interface RequestOptions {
 }
 
 export interface ResponseOptions {
-  status: number | MatcherResult
+  status: number
   headers?: { [name: string]: string | MatcherResult }
   body?: any
 }


TimothyJones:
Great catch, thank you!

TimothyJones:
Is there a reason this is marked as a draft, other than you wanted us to look at it before merging?

beckerei:
Sorry, just out of habit. But I'm not sure why some builds fail 👀

TimothyJones:
The build failures are an unrelated issue that only happens on PRs for some reason. Something to do with the git config.

I’ve merged it and will release first thing tomorrow. Thanks for the PR!

TimothyJones:
Releasing 9.17.1 now with this fix. Thanks again, and happy new year!

Issue #123, opened 2017-11-07
Please add --monkeypatch option to pact mock service start up command and pact-provider-verifier command


bethesque:
Requires standalone version pact-1.19.0.

This is to support: https://github.com/pact-foundation/pact-mock_service/issues/78 and another issue that I can't currently find.

The feature that the user is requesting is not something that should be changed for all users, as it is a non-standard behaviour, and breaks the Rack spec. However, we can allow the underlying code to be monkey patched for his specific use case. The value supplied to `--monkeypatch` will be an absolute path to a ruby file that will monkeypatch the underlying header logic.

This is not something that we need to document, but we can use it to deal with edge cases when necessary!

mefellows:
Moving this to `pact-node` as that is where the the CLI is integrated.

bethesque:
Thanks, I'll close this then.

PR #785, opened 2022-01-04
fix: All options are now passed down to pact-core in consumer pacts, allowing usage of the undocumented `monkeypatch` option

kevinkir:
This option can be used to monkey-patch the pact-mock-server. This addresses part of [issue #123](https://github.com/pact-foundation/pact-js/issues/123), which was originally closed due to the fact that option first had to be added to pact-node, which has since been done.

I am using jest-pact, and I want to be able to provide a monkeypatch, in order to increase the max URI length for the pact server, as it is [hard-coded to 2083 characters in WEBrick](https://github.com/ruby/webrick/blob/563b8309976065ba008430f2f1483d5bc3a713cf/lib/webrick/httprequest.rb#L446), but I want to increase that limit for my tests.

After this change, I can create a file webrick.rb
```rb
WEBrick::HTTPRequest.const_set('MAX_URI_LENGTH', 10000)
```

Then I can provide the monkeypatch option to jest-pact, to point to that file:
```js
import path from 'path';
import { pactWith } from 'jest-pact';

pactWith(
  {
    consumer: 'My consumer',
    provider: 'My provider,
    monkeypatch: path.join(__dirname, '../../monkeypatch/webrick.rb')
  },
  (provider) => { ... }
);
```

### DIFF
--- a/src/httpPact.spec.ts
+++ b/src/httpPact.spec.ts
@@ -35,6 +35,7 @@ describe("Pact", () => {
     spec: 2,
     cors: false,
     pactfileWriteMode: "overwrite",
+    monkeypatch: "/path/to/monkeypatch.rb",
   } as PactOptionsComplete
 
   before(() => {
@@ -140,6 +141,7 @@ describe("Pact", () => {
           spec: 2,
           cors: false,
           pactfileWriteMode: "overwrite",
+          monkeypatch: "/path/to/monkeypatch.rb",
         })
       })
     })
--- a/src/httpPact.ts
+++ b/src/httpPact.ts
@@ -8,7 +8,7 @@ import { isPortAvailable } from "./common/net"
 import logger, { traceHttpInteractions, setLogLevel } from "./common/logger"
 import { MockService } from "./dsl/mockService"
 import { LogLevel, PactOptions, PactOptionsComplete } from "./dsl/options"
-import { Server } from "@pact-foundation/pact-node/src/server"
+import { Server, ServerOptions } from "@pact-foundation/pact-node/src/server"
 import VerificationError from "./errors/verificationError"
 import ConfigurationError from "./errors/configurationError"
 
@@ -278,19 +278,9 @@ export class Pact {
 
   private createServer(config: PactOptions) {
     this.server = serviceFactory.createServer({
-      consumer: this.opts.consumer,
-      cors: this.opts.cors,
-      dir: this.opts.dir,
-      host: this.opts.host,
-      log: this.opts.log,
-      pactFileWriteMode: this.opts.pactfileWriteMode,
+      timeout: 30000,
+      ...this.opts,
       port: config.port, // allow to be undefined
-      provider: this.opts.provider,
-      spec: this.opts.spec,
-      ssl: this.opts.ssl,
-      sslcert: this.opts.sslcert,
-      sslkey: this.opts.sslkey,
-      timeout: this.opts.timeout || 30000,
-    })
+    } as ServerOptions)
   }
 }


mefellows:
Thanks. I think we were avoiding at making this option too visible but it makes sense for typed interfaces. I'll take a look when properly in front of a computer.

Just note it won't be preserved in the next major update as ruby will be gone.

TimothyJones:
Would it not be better to make this change in the ruby core?

kevinkir:
> Would it not be better to make this change in the ruby core?

Do you mean monkey-patching the WEBrick server in  pact-foundation/pact-mock_service, as opposed to making the consumer do the monkey-patch if they want to extend the URI length limit? That would be preferable from my perspective, but I figured it would be a tougher sell. I'd be happy to make that change instead if it would be accepted.

kevinkir:
> Just note it won't be preserved in the next major update as ruby will be gone.

Will the mock server be written in node.js after that point? If so, I think my issue would go away, as the limit there is 80kb. Is there an ETA for the next major update?

TimothyJones:
No, it will be based on the Rust core. Unfortunately, there isn't an eta, as we've run into some issues with linking FFI libraries on node 14+.

TimothyJones:
So this PR has two components - one adds the type to the TS type, and the other adds the passthrough of this property to pact-core.

I think in general we should pass all properties down to the core - since otherwise we'll need to extend the code each time we change a property.

The monkeypatch option is currently undocumented, so I don't mind that this PR doesn't add documentation. However, by the same token, I think perhaps we shouldn't have `monkeypatch` in the TS type - since otherwise we'll have an undocumented option appearing in user's intellisense? What do you think, @kevinkir?

If we do have it in the type, we could use `@deprecated` from jsdoc for intellisense, but we're not using jsdoc for the other properties either (which we should probably consider doing).

Also, commit messages prefixed with `feat` will appear in the changelog. Would you mind pushing a commit with a more appropriate commit message? You can find the guidelines for commit messages here: https://github.com/pact-foundation/pact-js/blob/master/CONTRIBUTING.md#release-notes

TimothyJones:
Also, please ignore the build failures on ubuntu. Something has changed in the way github does clones on PRs, I think.

kevinkir:
Thanks for the help @TimothyJones . I changed the commit to `refactor`, excluded `monkeypatch` from the types, and I modified `createServer` to pass down all the options. It passes down `logLevel` now, which it wasn't before. It looks to me like that won't have any effect, but if that is a concern, I can refactor to exclude it:

```js
 const { logLevel, ...serverOpts } = this.opts;
this.server = serviceFactory.createServer({
  timeout: 30000,
  ...serverOpts,
  port: config.port, // allow to be undefined
})
```

I don't have a preference between excluding the option from the types vs. marking it as `@deprecated`. Perhaps a small advantage of marking it as deprecated would be that it is a reminder to put this option's removal as a breaking change in the changelog in the version which migrates from the ruby server to the rust server? What do you think?

TimothyJones:
> Perhaps a small advantage of marking it as deprecated would be that it is a reminder to put this option's removal as a breaking change in the changelog in the version which migrates from the ruby server to the rust server?

I like your thinking - However, we actually already have this in this changelog and the migration guide - which will be copied in to pact-js when pact-js-core is bumped: https://github.com/pact-foundation/pact-js-core/blob/master/MIGRATION.md

Apologies, I think I wasn't clear in my feedback about the commit message - I think it should be included in the changelog, but with a more descriptive message. I'll squash merge it and update the message.

TimothyJones:
The release is cooking now, assuming it all goes well, this will be out as 9.17.2 shortly.

Thank you very much for the PR!

PR #802, opened 2022-01-20
Debug issue with the ruby standalone and git on PR requests

TimothyJones:
Not to be merged, just to peek under the hood of the interaction between github's checkout and the ruby standalone


YOU54F:
Thanks for this @TimothyJones.

Closing now as we have resolved the issue with the builds now!

PR #837, opened 2022-03-11
docs: Adjust README to make searching for properties easier

PauloGoncalvesBH:
## About

Whenever I needed to look for some property in the documentation to understand how I used it, I had a hard time finding it. First I did the search using the browser search (_ctrl+f_), then I had to go through the entire doc until I found the option to expand and see the table.
This is a behavior that has happened to me and I have seen it happen to at least 2 other people.

It is a more common situation to occur with people who are not fluent in English and who would like to read only the information about the `pactBrokerUsername` (example) property instead of having to translate the sections of the README to know where this property is possibly located.

I know this change will increase the size of the documentation, but it will make it easier to search for properties.

I understand the size of the documentation is not so important if it is fulfilling the role of passing the information.

### Before
image

### After
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Read [Getting started with Pact] for more information for beginners.
   - [HTTP API Testing](#http-api-testing)
     - [Consumer Side Testing](#consumer-side-testing)
       - [API](#api)
+      - [Constructor](#constructor)
       - [Example](#example)
     - [Provider API Testing](#provider-api-testing)
       - [Verification Options](#verification-options)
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks I think this is a fair point. I'd like to address the documentation in the next major release, but let's not stop that bringing this in.

May I ask - we do use TS under the hood which should help with understanding the API and available options. Do you use any IDE support to help and if so, why do these docs help (this will help us better understand where we can be focussed improving documentation etc. going forward)

Issue #372, opened 2019-10-04
Provider state parameters not implemented


kneczaj:
## Software versions

* **OS**: Windows 10
* **Consumer Pact library**:  @pact-foundation/pact@9.1.1
* **Provider Pact library**: does not matter
* **Node Version**: v10.15.3

## Expected behaviour

The state parameters can be passed, like at pact 3 specification
https://github.com/pact-foundation/pact-specification/tree/version-3#allow-multiple-provider-states-with-parameters

## Actual behaviour

A provider state can be defined by string only.

## Steps to reproduce

See the interface used to define an interaction that it can only receive a string for provider state
https://github.com/pact-foundation/pact-js/blob/3d66117c5b3ba0f2f15c9b29f5651c4eab1217cb/src/dsl/interaction.ts#L28

## Relevant log files

none

I am in a project which used to be consisted of Java services only which have defined many providers with parameters. With pact-js I cannot reuse these providers.

mefellows:
Thanks. We know this is lacking as it is a v3 feature. I'm going to create a separate issue later tonight to track the v3 implementation, but work is already underway (see the v3 branch).

kneczaj:
@mefellows thanks a lot!

straurob:
Is there any way to work around this with the v2 specification?

Example:
The consumer test registers a new user using her email address `foo@bar.com`. The provider state setup function must execute a query to make sure that the email address does not exist yet.

bethesque:
The better way to handle that solution is that the datasources should all be cleared at the start of the provider state set up call. Then you would know that it does not exist.

mefellows:
As Beth said. But also, you could have two separate tests and states

e.g.

* `"... when user foo@bar.com exists"`
* `"... when user foo@bar.com does not exist"`

You don't need a state parameter to do this.

garry-jeromson:
Are there any plans to make the library compatible with v3.0 state parameters? When you only have one variable, adding two tests is fine, but with several variables you end up needing N-factorial state resolvers on the provider side, which is really a pain.

For example, we have a multi-tenant application, and using [pactman](https://pypi.org/project/pactman/) we can structure our Python consumer tests to have states like:

```py
...
"providerStates": [
        {
          "name": "a tenant exists with the given tenant ID",
          "params": {
            "tenant_id": 1
          }
        },
        {
          "name": "a doctor with the given ID exists",
          "params": {
            "doctor_id": 123,
            "tenant_id": 1
          }
[...]

PR #882, opened 2022-06-30
fix: Add support for state params to MessageProviderPact (#372)

MichaelTamm:
This PR adds support for passing the params of providerStates to the `stateHandlers` of a `MessageProviderPact`.

### DIFF
--- a/src/dsl/message.ts
+++ b/src/dsl/message.ts
@@ -12,19 +12,29 @@ export interface Metadata {
 }
 
 /**
+ * Defines a state a provider must be in.
+ */
+export interface ProviderState {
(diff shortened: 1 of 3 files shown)


mefellows (approved):
Thanks!

FWIW this is still in the V2 part of the code base and I'm not actually sure that the core (Ruby) fully supports parameters.

The work in the new `pact-core` module exposes methods to do message based tests, but this code path still uses the Ruby core.

I'm not sure yet if this should stay backwards compatible with the previous Ruby core and have a new message interface exposed for the V3. But I'll bring this in for now and see how we go.

PR #883, opened 2022-07-05
chore: fix create mate tests

individual-it:
This fixes some tests. There are more issues in these tests, but for now fixing two issues:

- name tests and interactions
  (so that separate interactions are created)
- fix expected result for x-www-form-urlencoded data request
  (it only sends limited data, so it should only expect that as result)


mefellows (approved):
Thank you!

Issue #890, opened 2022-07-08
rename and upgrade @pact-foundation/pact-node@10.17.2 to @pact-foundation/pact-core@13.x.x


rbedemann:
### Checklist

Before making a feature request, I have: 

- [x] [Searched the issues to check that this feature hasn't been requested before](https://github.com/pact-foundation/pact-js/issues?q=is%3Aissue)
- [x] Checked the documentation to see if it is possible to do what I want already 

### Feature description

`@pact-foundation/pact-node` apparently got renamed in version [10.12.0](https://github.com/pact-foundation/pact-js-core/compare/v10.11.11...v10.12.0) to `@pact-foundation/pact-core`. `pact-js` currently uses version [10.17.2](https://github.com/pact-foundation/pact-js/blob/181d8a28f1aa55502d6103cadb6ae4b2c52f7ebf/package.json#L96) while version [13.6.2](https://github.com/pact-foundation/pact-js-core/releases/tag/v13.6.2) is already available. We'd love to see this package being renamed and upgraded.

### Use case
As we're currently facing some issue with our corporate proxy while downloading the pact binary (back in the days done via needle), we hope this version upgrade to the latest pact-core version could resolve this issue.

mefellows:
Thanks for submitting a PR @rbedemann, much appreciated.

mefellows:
See #891 for more.

PR #891, opened 2022-07-08
Migrate from @pact-foundation/pact-node to @pact-foundation/pact-core

rbedemann:
This fixes issue #890


mefellows:
Thanks for this @rbedemann and apologies for the delay - I was off traveling for work and am playing a bit of catch up. I've just approved the build to run.

FWIW `10.x.x` is now live and uses the new core. I'm not sure why we didn't integrate the newer package line into Pact JS a while back - I'll have a chat with my colleague to understand if there was a good reason. If this build passes though I see no reason not to release it into the `9.x.x` line so you don't need to upgrade to 10.x.x.

TimothyJones:
The build won't pass without substantial code changes, as upgrading brings (several) significantly breaking changes, some of which affect the exported surface from pact-js - so they would break the contract of the exported pact-js API. 

You _could_ bring it in to the 9.x.x line by following the instructions in the pact-js-core migration guide, but since it's a breaking change it would have to be released as 10 (and that's the reason that we did not upgrade it).

I would have preferred to bring it in and release a set of breaking changes without the v3 support branch, but we weren't able to do that without breaking everyone on the already released `10.0.0-beta` - since a release of `10.0.0` would match.

As an aside, if there is ever a need for a long lived pre-release in the future, I think it would be better if it were not tagged as a prerelease of the next major version. This caused a lot of pain.

Another aside, it would be good to put "please make sure the build and tests pass" in the [contributing guidelines](https://github.com/pact-foundation/pact-js/blob/master/CONTRIBUTING.md)

mefellows:
Thanks Tim. I assumed if it were this easy we'd have done it which I why I was so surprised to see the PR. 

Note the build didn't run until today as it needed a maintainer approval to unblock it.

TimothyJones:
You can run the build locally.

It’s possible it might have passed, when I first read this PR, for some reason I thought it was just the version bump and rename of imports. 

Anyway, the point remains that this is a breaking change and should not be merged into 9.x

TimothyJones:
I think this PR might illustrate that either:

* it’s worth merging pact-js and pact-js-core, 
or
* or the boundary between them isn’t quite right

It’s not ideal that a major version bump of an internal package can cause a major version bump of the user-facing package.

Now that pact-web is dead, it makes less sense to have a separate package. I previously was in favour of a separate CLI package, but really I can’t think of a case where you’d want pact-js but not the CLI.

mefellows:
Thanks Tim. I'm still unclear as to why it's a version bumping change from the PR. I'll have to look at the release notes between the `pact-core` changes to understand what the impact is.

@rbedemann if it's just avoiding the downloading of the standalone that's the issue, that can be resolved via two other means:

1. Upgrade to the 10.x.x line of Pact JS in your project (the HTTP interface is backwards compatible for the majority of use cases)
2. Modify the `pact-node` branch of the `pact-js-core` project to not install on download, instead bundling the Ruby binaries into it, and then merging that into Pact JS (which should then just be a minor version bump)

mefellows:
Also, if we are to merge this in, the base branch to merge into is now `9.x.x`.

TimothyJones:
> Thanks Tim. I'm still unclear as to why it's a version bumping change from the PR.

See https://github.com/pact-foundation/pact-js-core/blob/master/MIGRATION.md - most of these breaking changes to types and options are directly exposed by pact-js.

mefellows:
Thanks Tim.

That's right, I forgot we re-exported those options (probably not  great practice really!). 

> I think this PR might illustrate that either:
> it’s worth merging pact-js and pact-js-core,
> or
> or the boundary between them isn’t quite right
> It’s not ideal that a major version bump of an internal package can cause a major version bump of the user-facing package.

Yep. I think in this case it was probably going to be unavoidable given the large differences in the underlying engine, but the case for merging is definitely higher.

I'm going to close this off because doing that would require a major version bump, and the major version already has these changes in it.

Thanks @rbedemann for the effort. If you really need it in the 9.x.x line, the notes for that are above.

PR #905, opened 2022-08-03
chore: reference source file by relative path

lukiano:
### PR Template

Fix an issue where TS is looking for a module named `dsl/mockService` and fails to find it. By making the path relative it stops complaining.

### DIFF
--- a/src/httpPact/index.ts
+++ b/src/httpPact/index.ts
@@ -26,7 +26,7 @@ import { SpecificationVersion } from '../v3';
 import { version as pactPackageVersion } from '../../package.json';
 import { generateMockServerError } from '../v3/display';
 import { numberToSpec } from '../common/spec';
-import { MockService } from 'dsl/mockService';
+import { MockService } from '../dsl/mockService';
 import { setRequestDetails, setResponseDetails } from './ffi';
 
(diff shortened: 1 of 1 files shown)


mefellows:
Ah, thank you. The IDE often does this, I should see if there is a lint/rule that will catch this and prevent it.

I've just unlocked the build and if it passes we can merge and release.

lukiano:
Looks like the build failures are flukes (e.g. not able to connect to test.pactflow.io) but I don't know how to re-run them.

mefellows:
looks like a bad test setup, it's probably got the wrong selectors. I'll fix that - thanks!

YOU54F:
Thanks @lukiano / @mefellows - Noticed this last night when looking to update jest-pact to bring in support for v3 into the master branch

https://github.com/pact-foundation/jest-pact/runs/7642886870?check_suite_focus=true#step:4:128

one more to resolve re: express

YOU54F:
Just built `pact-js` locally from master, and linked to `jest-pact`, and re-ran the above GH actions run `npm run dist` and it has resolved the TS issue described by the OP 👍🏾 so this should be good to release

mefellows:
Thanks all!

PR #907, opened 2022-08-03
chore: npm old lockfile warning - update to 2.0

YOU54F:
Warning when running `npm i` on a clone of pact-js master branch
```
❯ npm i
npm WARN old lockfile 
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile 
```

I assume this should have the fix-up applied and committed, to avoid it repeating every time it is installed now

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,11417 @@
 {
   "name": "@pact-foundation/pact",
   "version": "10.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@pact-foundation/pact",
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks @YOU54F. I was having some build troubles a while back with the newer lockfile version (because we build across npm versions).

It looks like it's passing, so we can probably bring it in.

NOTE: there is an issue with the e2e test configuration for provider verification, which seems to occasionally not find any pacts to verify. I've got it on my list to look into.

PR #912, opened 2022-08-08
chore: update examples with latest pact-js (10.0.2) & jest-pact (0.10.1)

YOU54F:
Update to use latest version of jest-pact (0.10.1) which supports latest pact-js (10.0.2)

updates e2e examples to also use latest pact-js (10.0.2) and supporting lockfiles


mefellows:
Thanks Yousaf.

PR #913, opened 2022-08-09
docs: Correct async example

TimothyJones:
This fixes the new async example so that it will not always pass - although I think it's probably better to remove the wrapper function version entirely, as it's not necessary and makes it easier to make mistakes (like tests that always pass ;) ).


mefellows:
Thanks for this Tim. I have conceded (after attempting to write the tests in both Mocha and Jest) to remove the "wrapped" example, because in Jest it's definitely not necessary and Mocha it would be highly unusual, as it is usually paired with the `chai-as-promised` assertion library (mocha doesn't ship with an assertion library by default).

PR #914, opened 2022-08-11
feat: port extractPayload function from 9.x

goshatch:
The function is called `reify` in 10.x, and aliased as `extractPayload`
for backwards compatibility.

[Slack discussion](https://pact-foundation.slack.com/archives/C5F4KFKR8/p1660042113955029) for context.

### DIFF
--- a/docs/migrations/9-10.md
+++ b/docs/migrations/9-10.md
@@ -199,6 +199,9 @@ provider
     });
 ```
 
+#### Other
+* The `extractPayload` function on the matchers has been renamed `reify` in [v3] matchers. The old name (`extractPayload`) is still available as an alias, but might be deprecated in the future.
+
 [CLI]: https://docs.pact.io/implementation_guides/cli/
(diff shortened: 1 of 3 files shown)


YOU54F:
Thanks @goshatch 

Any chance you could add a line to the upgrade guide https://github.com/pact-foundation/pact-js/blob/master/docs/migrations/9-10.md

Would be good to get this added into the README/guide as well but we can support this but would be awesome if you wouldn't mind

goshatch:
Thanks @YOU54F! I've added a bit of documentation as you suggested. I'm happy to also update the README/guide, is this in a different repository?

mefellows:
Awesome thanks, this is great. We can add additional docs for the use case.

mefellows:
Once the build passes I'll merge and release

mefellows:
There's an issue with the PR pipeline unrelated to this. The core part of the build is good, merging and will release 🎉

PR #921, opened 2022-08-12
Add prepare script to allow installation of this package via GH

Maxim-Filimonov:
Based on https://github.com/npm/npm/issues/3055
npm introduced from v4 support for pre-compilation step on installing the package from node_modules or github directly. 

The reason this is needed is when people are trying to use their own forks of `pact-js` in their package.json directly. 
Right now, I hack it around with [special branch](https://github.com/Maxim-Filimonov/pact-js/tree/develop) which commits `dist` directory. But its a hack.

Its draft is that I don't want to put too much effort into figuring out how to test powershell(and sh) scripts **UNTIL** its confirmed that it this PR would be considered as valid change. 
They will need one extra addition - replace `'./dist'` with `''`.

@mefellows thoughts on this change, possible implications ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "postdist": "npm test",
     "predist": "npm run clean && npm run format:check && npm run lint",
     "release": "standard-version",
-    "test": "nyc --check-coverage --reporter=html --reporter=text-summary mocha"
+    "test": "nyc --check-coverage --reporter=html --reporter=text-summary mocha",
+    "prepare": "npm run compile"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
   "name": "@pact-foundation/pact",
   "version": "10.1.0",
   "description": "Pact for all things Javascript",
-  "main": "./src/index.js",
-  "types": "./src/index.d.ts",
+  "main": "./dist/src/index.js",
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks, I'll have to have a read of the source issue as looking at the changes I can't grok what's happening. I'll aim to take a look at this next week some time (or we can chat in slack as well for some more context).

mefellows:
Thanks @Maxim-Filimonov . I'm fine with supporting a `prepare` script this way, so long as the local development experience for the core maintainers doesn't change (or is at least as simple), and obviously, publishing a new package works ;)

TimothyJones:
What's in your fork? Would it be of general use?

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@pact-foundation/pact",
   "version": "10.1.0",
   "description": "Pact for all things Javascript",
-  "main": "./src/index.js",
-  "types": "./src/index.d.ts",
+  "main": "./dist/src/index.js",
```
TimothyJones, line 5: This won't work in the current state - the publish script actually creates an empty directory that copies in the dist folder, so the old path was correct.

Probably the right move is to update the publish script so it doesn't do that anymore - it was only a hack so that both pact-js and pact-web could be published from this repo, and pact-web isn't published any more.

mefellows:
Closing due to stale.

PR #929, opened 2022-08-24
chore: update examples dependencies

YOU54F:
- Updating example dependencies in v2/v3 folders
- Publish pacts with branch flag
- Update to use `absolute-version` as `@pact-foundation/absolute-version` is deprecated
- Update provider examples to publish with `providerVersionBranch` over deprecated `providerBranch`

It might be nice to split out the examples in the future, so they run as parallel jobs in CI, so we get both quicker feedback, and easier to pin point, and know a single example out of the lot failed


scripts/ci/build-and-test.sh
```diff
@@ -13,14 +13,15 @@ npm ci
 npm run dist
 cp package.json ./dist
 
-export GIT_BRANCH=${GIT_REF:11}
+export GIT_BRANCH=${GITHUB_HEAD_REF:-${GIT_REF#refs/heads/}}
```
YOU54F, line 16: Sets `GITHUB_HEAD_REF` for pull request synchronised builds

https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

> The head ref or source branch of the pull request in a workflow run. This property is only set when the event that triggers a workflow run is either pull_request or pull_request_target. For example, feature-branch-1.

Otherwise we get failures in the test runs, where the branch in a detached pull request build, doesn't match up, with the correct branch published with the Pact CLI (using `--auto-detect-version-properties`

mefellows: awesome, thanks - that's been on the todo-odle for a while!

PR #936, opened 2022-08-25
chore: update e2e examples

YOU54F:
Yo,

Not sure if there is an issue with the e2e example codebases, both the v2/v3 versions.

1. README states hit the consumer service with a bearer token https://github.com/pact-foundation/pact-js/blob/master/examples/e2e/README.md#matching-service
1. The consumer codebase doesn't set an authHeader in the request to the provider codebase, so fails the request https://github.com/pact-foundation/pact-js/blob/e7d71e45a4a286c40fdf1ad8b49c6377da2540dc/examples/e2e/consumer.js#L77
1. The pact provider tests verify, because if an Authorization token is set, it is switched out with the correct token https://github.com/pact-foundation/pact-js/blob/master/examples/e2e/test/provider.spec.js#L31

In order to get this to work E2E,  I have 

1. Updated consumer codebase to send correct token
1. Added regex matchers to Auth header, so when correct header is sent, an example is encoded into the pact file, avoiding the user serialising sensitive data into the pact
1. Updated the README so if calling the consumer service, no auth header is required to be set.

Consumer pact tests are generated, provider is able to verify, and application works e2e when run.

Not sure if it is meant to be in this state, if so, we can add documentation to explain why it doesn't work e2e and the steps to resolve

### DIFF
--- a/examples/e2e/README.md
+++ b/examples/e2e/README.md
@@ -104,7 +104,7 @@ curl -H "Authorization: Bearer 1234" -X POST -H "Content-Type: application/json"
 #### GET /suggestions/:id
 
 ```
-curl -H "Authorization: Bearer 1234" -X GET http://localhost:8080/suggestions/1
+curl -X GET http://localhost:8080/suggestions/1
 ```
 
 ## Viewing contracts with the Pact Broker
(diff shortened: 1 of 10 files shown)


mefellows:
I think it's just a case of drift over time and whilst the examples run, not all of the code is exercised.

I've been thinking we should create a proper set of examples to exercise the functionality we want, with a consistent domain/use case and with less arbitrary use of the various features.

Separate chat, but the a BDD specification seems like a good way to achieve this!

mefellows:
(also this is good to go if it is 🟢 )

Issue #884, opened 2022-07-05
[V3] body is modified to JSON object even for non-JSON data (regression in 10.0.0-beta.60)


individual-it:
### Software versions

_Please provide at least OS and version of pact-js_

- **OS**: Ubuntu 21.10
- **Consumer Pact library**: Pact JS v10.0.0-beta.60
- **Provider Pact library**:  Pact JS v10.0.0-beta.60
- **Node Version**: 16.0.0

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/#test-fails-when-it-should-pass) the section on intermittent test failures
- [ ] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [x] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour

when using `Content-Type: application/x-www-form-urlencoded` the data in the body would be send as it is

### Actual behaviour

data is encoded into a JSON object
an interaction like 
```
"description": "as 'admin', a POST request to create a user",
"request": {
  "body": "password=1234&userid=testUser1",
  "headers": {
    "Content-Type": "application/x-www-form-urlencoded",
    "authorization": "Basic YWRtaW46YWRtaW4="
  },
.....
```
results in this request to the provider
![image](https://user-images.githubusercontent.com/2425577/177291586-e9b59356-0af6-48a3-a263-ba6bd66c4c45.png)

### Steps to reproduce

1. create an interaction with `"Content-Type": "application/x-www-form-urlencoded"`
2. set a body
3. run provider tests
4. observe what is send to the provider

reproducible example: https://github.com/pact-foundation/pact-js/pull/883/files
the test 'creates a new mate with application/x-www-form-urlencoded data' works with `10.0.0-beta.59` but fails with `10.0.0-beta.60` because the data is not urlencoded but wrapped into a JSON object

mefellows:
It seems like this is working now @individual-it - are you able to confirm? The tests with that PR are definitely green, and it looks like the body is correct.

YOU54F:
So it looks like the test was failing, but `enablingPending:true` is set, so the build doesn't fail, where there are provider verification failures.


This is a build from master showing the failing verification from a build log

[...]

Issue #662, opened 2021-05-10
[V3] eachKeyLike doesn't ignore the key's value when matching


francislobo:
### Software versions

- **OS**: Win 10
- **Consumer Pact library**:  "@pact_foundation/pact": "10.0.0-beta.34", "jest-pact": "0.9.0-beta.v3",  
- **Provider Pact library**:  pact-jvm-provider-gradle', version: '4.0.10'
- **Node Version**: v12.13.1

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code
- [ ] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [x] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour

For an example data [shown here](https://gist.github.com/francislobo/329d85e843ab00f49840d7d1250785f1), the PACT generated using PACT JVM and PACT JS are different.
Specifically [this section](https://gist.github.com/francislobo/e85f651bb7dabe759f310c3b387c691b#file-theconsumer-theprovider-json-L43)generated from the eachkeylike matchers. This breaks the eachkeylike matcher where the actual key should be ignored and only the value (contents) of the key must be matched.

[Pact generated from JS](https://gist.github.com/francislobo/9a78a9939bfd0a66e690df80b76c196a)
[Pact generated from JVM](https://gist.github.com/francislobo/e85f651bb7dabe759f310c3b387c691b)  
NOTE: there are some extra pattern matchers on the request path of JVM PACT. I hadn't added them on the JS part as I was just getting started (I assume that shouldnt change how eachKeyLike works)

[JS code used to generate the contract](https://gist.github.com/francislobo/9f519c13973dc5b9b1edaf8163519c12)

### Actual behaviour

The PACTs should be identical for the eachkeylike section

### Steps to reproduce

_How can someone else reproduce this bug?_
The gist files has all the code required to reproduce this

### Relevant log files
none

[Slack conversation pertaining to this issue](https://pact-foundation.slack.com/archives/C9VBGLUM9/p1620164070259800)

mefellows:
Thanks @francislobo, I'll take a look at this on Monday.

YOU54F:
Just noted that this is still an issue

Failing verification

[...]

PR #937, opened 2022-08-25
chore (examples): update e2e example matchers

YOU54F:
Rectifies a few failures in provider verifications for the e2e example projects

1. Few tests would fail if they got an array with more than 1 item in, despite all objects matching the same shape, so wrapped in an `eachLike`
2. Set auth header on request to POST `/animals` on consumer side

Also added a cookie trail regarding 2 defects which were already raised, and present in our examples.

https://github.com/pact-foundation/pact-js/issues/884
https://github.com/pact-foundation/pact-js/issues/662

Failures on verification don't fail the test builds for the `e2e` and `v3/e2e` projects as `enablePending: true`


mefellows:
Thanks!

Will bring this in asap. I've just pushed up a fix for the example testing.

mefellows:
Thanks!

Will bring this in asap. I've just pushed up a fix for the example testing.

PR #940, opened 2022-08-31
fix: Add Date as a valid JSON value

jwmcconnell:

## PR Template

Add Date as a valid JSON value.

### DIFF
--- a/src/common/jsonTypes.ts
+++ b/src/common/jsonTypes.ts
@@ -1,4 +1,11 @@
-export type AnyJson = boolean | number | string | null | JsonArray | JsonMap;
+export type AnyJson =
+  | boolean
+  | number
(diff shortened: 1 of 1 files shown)


mefellows:
Unless I'm mistaken, Date is not a valid JSON value type. Can you please elaborate on what you're trying to achieve here?

TimothyJones:
I agree with Matt.

jwmcconnell:
> Unless I'm mistaken, Date is not a valid JSON value type. Can you please elaborate on what you're trying to achieve here?

Sorry for the confusion, I was attempting to transform an example of yours to use one of my types and made a mistake with the consumer test.

mefellows:
No probs!

PR #942, opened 2022-09-06
fix: PactV4 error handling logic

mefellows:
See repro here: https://github.com/lennartquerter/pact-v3-test

### DIFF
--- a/src/v3/pact.ts
+++ b/src/v3/pact.ts
@@ -199,8 +199,6 @@ export class PactV3 {
     try {
       val = await testFn(server);
     } catch (e) {
-      this.cleanup(false, server);
-
       error = e;
     }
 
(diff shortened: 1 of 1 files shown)


TimothyJones:
Good find and fix. Sorry if this was my fault.

This is a good candidate for sticking to the [changelog guidelines](https://github.com/pact-foundation/pact-js/blob/master/CONTRIBUTING.md#release-notes) - the issue with developer-centric commits is that the [changelog for this release](https://github.com/pact-foundation/pact-js/releases/tag/v10.1.3) isn't very friendly for users, since it just says "PactV4 error handling logic". 

I reckon something like `fix: Correct an issue that caused certain tests to fail with "Error in native callback"` would have been better.

PR #946, opened 2022-09-21
fix: Fix issue where MatchersV3.string() with no arguments would not correctly match strings

### LINKED ISSUE #945: MatchersV3.string() with no arguments doesn't match strings
### Software versions

- **Consumer Pact library**: 10.1.3

### Issue Checklist

Please confirm the following:

- [X] I have upgraded to the latest
- [X] I have the read the FAQs in the Readme
- [X] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures

### Expected behaviour

`MatchersV3.string()` should match any string

### Actual behaviour

[this line has no default argument](https://github.com/pact-foundation/pact-js/blob/master/src/v3/matchers.ts#L256), so `MatchersV3.string()` actually produces `{ 'pact:matcher:type': 'type', value: undefined }` which will not work as expected.

### Steps to reproduce

Make a pact expectation that includes `MatchersV3.string()`

### Relevant log files
 
N/A

TimothyJones:

### PR Template

Fixes #945 

I also updated package-lock, since no current version of node uses the old package-lock format.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,11418 @@
 {
   "name": "@pact-foundation/pact",
   "version": "10.1.3",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@pact-foundation/pact",
+      "version": "10.1.3",
+      "license": "MIT",
+      "dependencies": {
+        "@pact-foundation/pact-core": "^13.7.8",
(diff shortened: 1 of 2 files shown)


mefellows:
I dunno man, where are the tests for this one? ;)

TimothyJones:
Who tests the testers?

PR #947, opened 2022-09-21
build: No longer remove docs folder in clean step

TimothyJones:

### PR Template

When preparing #946 I noticed that the `docs` folder was deleted following `npm run dist`. This is because they're deleted in the `clean` step - but, I don't see the docs being autogenerated anywhere.

My guess is that `docs` was left around from a previous iteration, and now that there is a docs folder again, it's not supposed to be removed by the build script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "types": "./src/index.d.ts",
   "scripts": {
     "compile": "rimraf dist && tsc",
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks! It used to be used for the `jsdoc` command - but we don't use/publish that anymore.

PR #953, opened 2022-09-28
docs: providerVersionBranch in provider opts

YOU54F:
Draft - start of a bigger change to update to reference new consumerVersionSelectors for mainBranch/deployedOrReleased and new providerVersionBranch option


mefellows:
Thanks!

Should we leave the old parameter there, with a note saying it's not recommended and the new one is?

YOU54F:
maybe, it wasn't in the docs anyway so I reckon probably not 🤷🏾‍♂️

docs/provider.md
```diff
@@ -60,7 +60,7 @@ new Verifier(opts).verifyProvider().then(function () {
...
 | `consumerVersionSelectors`  | false     | ConsumerVersionSelector\|array | Using [Selectors](https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors/) is a way we specify which pacticipants and versions we want to use when configuring verifications. |
 | `consumerVersionTags`       | false     | string\|array                  | Retrieve the latest pacts with given tag(s)                                                                                                                                                        |
-| `providerVersionTags`       | false     | string\|array                  | Tag(s) to apply to the provider application                                                                                                                                                        |
+| `providerVersionTags`       | false     | string\|array                  | Tag(s) to apply to the provider application                                   | `providerVersionBranch`       | false     | string                 | Branch to apply to provider application version (recommended to set) |
```
TimothyJones, line 63: I think this needs to be a new line on its own?

TimothyJones:
I don't actually know exactly what `--auto-detect-version-properties` does, but my thinking is that it should probably be the default unless it is overridden. It would be great if Pact just handled the versioning for you, since so many people just hard code a version and then wonder why it doesn't work.

docs/provider.md
YOU54F: It does, fail edit in the GH editor

YOU54F:
This is good for merging now, and I'll pick up the other updates across the individual libraries as part of this work

https://github.com/pact-foundation/docs.pact.io/pull/173

as we have code blocks for using branches/consumer version selectors when verifying on the website, and some of the language impls are lacking.

we can then think about reducing the cognitive load on the user for needing to set branch and version unless they explicitly set them

TimothyJones (approved):
Looks good to me!

For bonus points you could link to something that says why it’s recommended- but those bonus points shouldn’t stop you merging this as-is 🙌

YOU54F:
> Looks good to me!
> 
> For bonus points you could link to something that says why it’s recommended- but those bonus points shouldn’t stop you merging this as-is 🙌

TY homie

Will link back to the updated pages in docs.pact.io (in the CI/CD workflow and pact nirvana) when they are complete, and try and do it in a consistent way across repos

PR #970, opened 2022-11-08
fix reify logic relating to isMatcher function

hborham:
isMatcher inadvertantly passes when a field name is 'value'

Thank you for making a pull request!

Pact-JS is built and maintained by developers like you, and we appreciate contributions very much. You are awesome!

Here is a short checklist to give your PR the best start:

_Everything above can be removed once checked_

- [ ] Commit messages are ready to go in the changelog (see below for details)
- [ ] PR template filled in (see below for details)

# Commit messages

Our changelog is automatically built from our commit history, using conventional changelog. This means we'd like to take care that:

- commit messages with the prefix `fix:` or `fix(foo):` are suitable to be added to the changelog under "Fixes and improvements"
- commit messages with the prefix `feat:` or `feat(foo):` are suitable to be added to the changelog under "New features"

If you've made many commits that don't adhere to this style, we recommend squashing 
your commits to a new branch before making a PR. Alternatively, we can do a squash
merge, but you'll lose attribution for your change.

For more information please see CONTRIBUTING.md

_Everything above can be removed_

### PR Template

_Please describe what this PR is for, or link the issue that this PR fixes_

_You may add as much or as little context as you like here, whatever you think is right_

_Thanks again!_

### DIFF
--- a/src/v3/matchers.spec.ts
+++ b/src/v3/matchers.spec.ts
@@ -514,7 +514,7 @@ describe('V3 Matchers', () => {
   describe('#reify', () => {
     describe('when given an object with no matchers', () => {
       const object = {
-        some: 'data',
+        value: 'data', // field name of value should not be only determination of isMatcher()
         more: 'strings',
         an: ['array'],
         someObject: {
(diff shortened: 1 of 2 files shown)


hborham:
I suspect there's a preferred way to enhance the isMatcher to not be true for just AnyTemplate with a value field.  Please feel free to suggest and/or modify this PR

TimothyJones:
This is a great catch. I don't know why I didn't write it that way to start with - your way is definitely both more accurate and better. 

I'm not a maintainer any more so I can't enable the workflow - but it would be great if you could push a different commit message that adheres to [the guidelines](https://github.com/pact-foundation/pact-js/blob/master/CONTRIBUTING.md#commit-messages) - it needs a `:` after `fix` so that it gets picked up by the tooling, and it needs to be a message that makes sense in the changelog. Something like `fix: Fix an issue where extractPayload would not work correctly with objects with a value key`

Thanks for covering it with a test too! Much appreciated!

hborham:
Thanks - just amended.

TimothyJones (approved):
Nice work! Thanks for the PR

mefellows:
Agreed 🙏

Issue #974, opened 2022-11-14
Getting `Cannot find module 'dsl/message'` error in 10.2.0


qmg-akara:
Hi guys, thanks for the great work!

We just upgraded to version 10.2.0 but our `yarn lint` is failing due to 2 import paths:
```
node_modules/.store/@pact-foundation-pact-npm-10.2.0-8ecf80a526/node_modules/@pact-foundation/pact/src/dsl/verifier/proxy/types.d.ts:4:34 - error TS2307: Cannot find module 'dsl/message' or its corresponding type declarations.

4 import { MessageProviders } from 'dsl/message';
                                   ~~~~~~~~~~~~~

node_modules/.store/@pact-foundation-pact-npm-10.2.0-8ecf80a526/node_modules/@pact-foundation/pact/src/dsl/verifier/types.d.ts:2:40 - error TS2307: Cannot find module 'dsl/options' or its corresponding type declarations.

2 import { MessageProviderOptions } from 'dsl/options';
                                         ~~~~~~~~~~~~~
```

Are aliased import paths typo or intentional?

### Software versions

_Please provide at least OS and version of pact-js_

- **OS**: _e.g. Mac OSX 10.11.5_
- **Consumer Pact library**: _Pact JS v10.2.0_
- **Provider Pact library**: _N/A_
- **Node Version**: `18`

### Issue Checklist

Please confirm the following:

- [X] I have upgraded to the latest
- [ ] I have the read the FAQs in the Readme
- [ ] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures
- [ ] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [ ] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour

No linting error

### Actual behaviour

Getting the above errors

### Steps to reproduce

It can be tested with a fresh install

### Relevant log files

See above

mefellows:
Gah! I really need to find a way to prevent my editor from using relative imports, and investigate if this can be easily added to our CI to fail a build.

I'll get a patch out ASAP.

(also, thank you for the kind words)

qmg-akara:
Thanks for the quick fix!

mefellows:
You're welcome

TimothyJones:
[...]

PR #975, opened 2022-11-14
Fix/prevent absolute imports

mefellows:
This change detects any use of absolute paths such as `dsl/message`, resulting in a compilation failure.

VS Code (and hopefully other IDEs) will not auto-suggest/import these paths, and provide helpful hints such as the following if they are used:

Screen Shot 2022-11-14 at 10 33 11 pm

See also #974


mefellows:
The Windows build failure is unrelated, it seems the recent "fix" for that did not stick 🤔 .

PR #1008, opened 2022-12-05
fix: Remove unnecessary dependency @types/bluebird (also removed bluebird internally)

TimothyJones:

### PR Template

This PR removes an unnecessary `@types/bluebird` dependency. When looking in to whether or not Pact is actually using bluebird, I noticed that it _was_, but that it wasn't necessary. So I removed it.

This PR replaces Bluebird with native promises.

There is one small behaviour change - this change will no longer run the localhost port checks sequentially. I don't think this is actually necessary, since we're only checking if any one of them passes, it won't matter if `127.0.0.1` and `localhost` collide.

Also, I noticed the tests are hanging at `TRACE (72171): pact@10.3.1: Verification completed, closing server` for about 30 seconds. I don't know why this is, but it was happening in master too, so I ignored it.

### DIFF
--- a/src/common/net.ts
+++ b/src/common/net.ts
@@ -31,9 +31,7 @@ export const isPortAvailable = (port: number, host: string): Promise<void> =>
   ).then((settledPortChecks) => {
     // if every port check failed, then fail the `isPortAvailable` check
     if (settledPortChecks.every((result) => result.status === 'rejected')) {
-      return Promise.reject(
-        new Error(`Cannot open port ${port} on ipv4 or ipv6 interfaces`)
-      );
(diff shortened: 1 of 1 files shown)


TimothyJones:
Looks like this build is failing for the same reason that master is failing. I'll wait for a maintainer to look in to that 😝

mefellows:
Thanks Tim. I brought in a bunch of the dependency updates to the examples, but looks like it 💥 .

They are getting a little unwieldy now and I've been thinking if we need to take a different route (the majority of build time now is running these "integration" examples, and I'm becoming increasingly convinced there is probably a better path forward here).

But that aside, this is much appreciated and makes sense - thanks!

PR #1009, opened 2022-12-05
fix: Export `VerifierOptions` as a root export

TimothyJones:

### PR Template

While looking at https://github.com/pact-foundation/nestjs-pact/pull/17  , I noticed that one of the breaking changes in V10 was that `VerifierOptions` is now deeply imported. This PR fixes that by exporting it at the root again.


mefellows:
Thanks!

PR #1027, opened 2022-12-20
Remove most lint directives

TimothyJones:
This PR removes almost all lint directives from the source. Each commit is reasonably self-contained, so you could cherry-pick if you don't want some of them.

Key points:

* The typescript type checker was explicitly turned off (!) in a few files. This is not a good idea, as it means that land mines and maintainability issues are left for future developers
* Some private properties are relied on in tests. Since most of these were on the new data classes, I just made the data class properties public.
* Some tests either used incomplete mocks, or wrote to private properties. I left those with TODO comments
* I didn't touch V4 because it looks like it's work in progress.
* The javascript eslint settings are not disabled by the corresponding typescript rules for some reason. This is annoying, as it means there are two settings for the same rule in the rules file. This should be fixed, but it's out of scope for this PR.
* Almost all lint directives were unnecessary

### DIFF
--- a/src/dsl/verifier/proxy/tracer.ts
+++ b/src/dsl/verifier/proxy/tracer.ts
@@ -45,9 +45,6 @@ export const createResponseTracer =
       return oldWrite.apply(res, [chunk]);
     };
 
-    // I think the type definitions in @types/node is wrong/broken.
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
     res.end = (chunk: Parameters<typeof res.write>[0]) => {
       if (chunk) {
         chunks.push(Buffer.from(chunk));
@@ -58,7 +55,7 @@ export const createResponseTracer =
           removeEmptyResponseProperties(body, res)
         )}`
       );
-      oldEnd.apply(res, [chunk]);
+      return oldEnd.apply(res, [chunk]);
     };
     if (typeof next === 'function') {
       next();
--- a/src/dsl/verifier/verifier.ts
+++ b/src/dsl/verifier/verifier.ts
@@ -2,19 +2,20 @@
  * Provider Verifier service
  * @module ProviderVerifier
  */
-import serviceFactory from '@pact-foundation/pact-core';
-import { VerifierOptions as PactCoreVerifierOptions } from '@pact-foundation/pact-core';
+import serviceFactory, {
+  VerifierOptions as PactCoreVerifierOptions,
+} from '@pact-foundation/pact-core';
 import { omit, isEmpty } from 'lodash';
 import * as http from 'http';
 import * as url from 'url';
 
+import { AddressInfo } from 'net';
 import logger, { setLogLevel } from '../../common/logger';
 
 import ConfigurationError from '../../errors/configurationError';
 import { localAddresses } from '../../common/net';
 import { createProxy, waitForServerReady } from './proxy';
 import { VerifierOptions } from './types';
-import { AddressInfo } from 'net';
 
 export class Verifier {
   private address = 'http://127.0.0.1';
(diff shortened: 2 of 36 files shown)


TimothyJones:
Update: I fixed the duplicate rules problem by using the appropriate typescript lint settings. This brought in some new rules about dependency cycles and use-before-define, so I fixed those too. I don't think they're actually problems with the current versions of node, but it's still good to be standards compliant.

Also, I removed eslint-plugin-prettier since it wasn't used (pact uses the config instead). Using both is a common mistake - for those playing along at home:

* the plugin is to make the linter do prettier formatting (Pact did not do this)
* the config is to stop the linter conflicting with prettier (Pact did do this)

Using the config (with prettier run separately on save) is a nicer developer experience because it means that you only get red squiggles in the editor if you're typing code that isn't right, rather than whenever the formatter hasn't been run.

src/dsl/verifier/proxy/tracer.ts
```diff
@@ -58,7 +55,7 @@ export const createResponseTracer =
           removeEmptyResponseProperties(body, res)
         )}`
       );
-      oldEnd.apply(res, [chunk]);
+      return oldEnd.apply(res, [chunk]);
```
mefellows, line 58: 🤦

src/dsl/verifier/verifier.ts
```diff
@@ -119,14 +120,14 @@ export class Verifier {
...
-      const port = (server.address() as AddressInfo).port;
+      const { port } = server.address() as AddressInfo;
       const opts: PactCoreVerifierOptions = {
         providerStatesSetupUrl: `${this.address}:${port}${this.stateSetupPath}`,
         ...omit(this.config, 'handlers'),
         providerBaseUrl: `${this.address}:${port}`,
         transports: this.config.transports?.concat([
           {
-            port: port,
+            port,
```
mefellows, line 130: I remember a time that the eslint plugin would just do these types of changes on save. I'm assuming that's what's happening in, in which case - huzzah!

src/httpPact/index.spec.ts
```diff
@@ -154,10 +154,12 @@ describe('Pact', () => {
         withResponseBody,
         withResponseHeader,
         withStatus,
-      };
+      } as unknown as ConsumerInteraction; // TODO replace with proper mock
+      // @ts-ignore TODO refactor the class to remove the need for this
       p.pact = pactMock;
+      // @ts-ignore: TODO refactor the class to remove the need for this
```
mefellows, line 160: +1

src/httpPact/tracing.ts
```diff
@@ -4,13 +4,20 @@ import logger from '../common/logger';
...
-  // TODO: Need to look into the types here. They seemed to have changed
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
   http.request = (
-    options: RequestOptions,
-    cb: (res: IncomingMessage) => void
+    options: RequestOptions | string | URL,
+    cb: RequestOptions | ((res: IncomingMessage) => void) | undefined
   ): ClientRequest => {
+    if (typeof options === 'string' || options instanceof URL) {
```
mefellows, line 11: 👍

mefellows (approved):
:chefkiss: (<- they don't have this emoji, but I will continue to use it until they do)

mefellows:
> Update: I fixed the duplicate rules problem by using the appropriate typescript lint settings. This brought in some new rules about dependency cycles and use-before-define, so I fixed those too. I don't think they're actually problems with the current versions of node, but it's still good to be standards compliant.
> 
> Also, I removed eslint-plugin-prettier since it wasn't used (pact uses the config instead). Using both is a common mistake - for those playing along at home:
> 
> * the plugin is to make the linter do prettier formatting (Pact did not do this)
> * the config is to stop the linter conflicting with prettier (Pact did do this)
> 
> Using the config (with prettier run separately on save) is a nicer developer experience because it means that you only get red squiggles in the editor if you're typing code that isn't right, rather than whenever the formatter hasn't been run.

Right, this explanation made sense. I think in the early days, we must have had some eslint rules that were constantly fighting with prettier. I'm pretty sure I introduced prettier, and this was the solution I ended up with.

But what I really want is this:

* lint - find bugs, don't concern itself with styling
* prettier - make the code look the same, no matter who writes it. Remove arguments about opinions (except semicolons of course 😆 )

TimothyJones:
> But what I really want is this:
> * lint - find bugs, don't concern itself with styling
> * prettier - make the code look the same, no matter who writes it. Remove arguments about opinions (except semicolons of course 😆 )

I agree. This is the current state of pact linting (except for nestjs-pact, which has the conventions it was written with before it moved to pact-foundation)

The defaults for prettier include semicolons, of course ;)

PR #1034, opened 2022-12-23
build: Improve logging in install-plugins script

TimothyJones:
This PR:

* Improves the logging for when install-plugins fails
* Ensures that windows actually fails when install-plugins fails

It exposes the problem with install-plugins, which is that it is being rate limited by github. You could fiddle about with a token to fix this, but probably it would be better to do one of:

* Pin the version, avoiding the need for the call (easiest)
* Replace the API call with https://github.com/marketplace/actions/github-api-request
* Remove the need for a plugin cli step entirely by using a package manager (imo, best)

### DIFF
--- a/scripts/ci/build-and-test.sh
+++ b/scripts/ci/build-and-test.sh
@@ -6,14 +6,13 @@ set -x
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)" # Figure out where the script is running
 . "$SCRIPT_DIR"/lib/robust-bash.sh
 
-require_env_var GIT_REF
 
 npm ci
 
 npm run dist
 cp package.json ./dist
 
-export GIT_BRANCH=${GITHUB_HEAD_REF:-${GIT_REF#refs/heads/}}
+export GIT_BRANCH=master
 
 export PACT_BROKER_USERNAME="dXfltyFMgNOFZAxr8io9wJ37iUpY42M"
 export PACT_BROKER_PASSWORD="O5AIZWxelWbLvqMd8PkAVycBJh2Psyg1"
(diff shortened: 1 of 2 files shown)


TimothyJones:
I've also got a branch that pushes the `GITHUB_TOKEN` down to that curl script. You can find it here:

https://github.com/TimothyJones/pact-js/tree/use-token-in-install-plugins

the relevant commit is https://github.com/TimothyJones/pact-js/commit/7654bc50e43ffc8a7fc0038b1ea844e7aab937b8

I didn't include it in this PR because it seems unnecessary to pass the token around just to download a dependency. But it appears to fix this brittle build problem.

YOU54F:
Thanks for this Tim

I think pinning the version would be nice, we could have a json file that stores the name of the plugin and the version.

On release of a new plugin, it could trigger a script to update the json metadata triggering a build.

Would allow for us to create and test beta versions of particular plugins easily, e2e

YOU54F:
I'm tempted to hard-code/pin the tag for now to `0.0.4` for the time being, so we can yak shave through the rest of these PR's, no-one likes a flaky build!

mefellows (approved):

mefellows:
Thanks Tim!

> Replace the API call with https://github.com/marketplace/actions/github-api-request

That's a good idea, I wasn't aware of that!

> Remove the need for a plugin cli step entirely by using a package manager (imo, best)

We created initial support for Plugins this way which gives us a lot of flexibility as we evolve. This (native language packaging) would possibly be the best developer experience, but it would then mean every plugin author has to distribute via npm, rubygems, Maven, Nuget etc. etc. 

There's nothing stopping us, say, creating a plugin config file or extending package.json with defined plugins, and having the framework discover and download them. We looked at several plugin models at the time, and decided this was a good starting point. As we collect feedback from the community we can adjust and improve (or pivot, as the case may be) our approach.

mefellows:
I'll merge, and then pin for now. We can improve the solution as the need arises.

PR #1035, opened 2022-12-24
chore: Update dependencies to fix #1029

### LINKED ISSUE #1029: Review the dependencies and remove unused ones
It looks like there are several dependencies that aren't used or could be removed. For example, in devDependencies, I don't think `enhanced-resolve` or `jasmine-core` are used. In the main dependencies, I don't think `pino-pretty` is used (if it's a peer-dep that's needed, it should be in pact-js-core). There are probably others - these are just the ones I noticed when fixing some lint issues.

As a bonus, almost all uses of lodash and ramda can be replaced with native TS code now - both those libraries are unnecessarily large and were the source of many of the vulnerability warnings in the past.

TimothyJones:
This PR:

* Fixes #1029 by removing (as far as I can tell) unused dependencies
* Runs audit fix on all examples that currently pass without changes
* Substantially improves the run audit fix on examples script to:
   * Keep going when an example fails, since the others might pass 
   * Also print out the results of npm audit
   * Better separate the examples in the log
   * Print a list of all the failed updates at the end

Screenshot attached:
Screen Shot 2022-12-24 at 3 46 50 pm


TimothyJones:
The tests pass locally. However, the issue highlighted by https://github.com/pact-foundation/pact-js/pull/1034 is still present, so we can't see the output from the build.

mefellows:
Thanks Tim! I'm wary of the audit changes with respect to the examples (build failures unlikely to relate to PRs), however, this improves the current situation so worth doing for now until we find a better solution to the example problem.

TimothyJones:
I think the examples just need to be maintained. Probably most of them can be killed- like, nowadays you could just do typescript for all, which would kill the need for the “typescript” standalone example. Making all of them end to end would also remove the need for “e2e”

PR #1036, opened 2022-12-28
Other cleanup

TimothyJones:
Fixing a couple of issues I noticed that might be problems later.


mefellows:
Thanks. It looks like a few of those updates were in another branch I merged. I've manually resolved them but I expect prettier to get upset at something.

PR #1056, opened 2023-02-07
docs: Add a note for Alpine + Docker

drakulavich:
The PR adds a workaround on how to install `pact-js` using a base image `node:18-alpine` for example.

It might be helpful when you use Alpine for production build and you want just install the dependencies without actual pact tests execution.


mefellows:
Nice - thank you!

PR #1061, opened 2023-02-16
chore: Improve types

### LINKED ISSUE #1054: Question: How to use TypeScript interfaces with matchers
Hello,

this is a question rather than a bug report or feature request.

I'm quite new to Pact and I would like to reuse the TypeScript interfaces my frontend app uses for the consumer-side of things.

The [documentation](https://docs.pact.io/implementation_guides/javascript/docs/matching#a-note-about-typescript) about matching states that in order to use TypeScript interfaces inside e.g. `willRespondWith/body` you need to wrap them in `InterfaceToTemplate` or use `type` instead of interface.

I've tried both but the results where not quite satisfactory.

As far as I understand it would be beneficial to use existing `type`s and `interface`s to prevent typos and to define the correct expected value types for the mock server.

On the other hand, `Matcher.<something>` seems incompatible with the types expected by interface members. The example from the docs only uses the `like` matcher _around_ the interface, but not for its members.

My question is how I would be able to achieve something like this with Pact:

```typescript
interface Foo {
  a: string;
}

const f: InterfaceToTemplate<Foo> = { a: like("working example") };

// The above causes:
// Type 'Matcher<"working example">' is not assignable to type 'AnyTemplate'.
//    Type 'Matcher<"working example">' is not assignable to type 'TemplateMap'.
//       Index signature for type 'string' is missing in type 'Matcher<"working example">'.

provider.addInteraction({
  uponReceiving: "a post with foo",
  withRequest: {
    method: "POST",
    path: "/",
    body: like(f)
  },
  ...
})
```

The only thing that comes to mind is to slap `as unknown as string` after the matcher.

### Software versions

- **OS**: macOS Ventura
- **Consumer Pact library**: `Pact JS 10.4.1`
- **Node Version**: `v16.19.0`

TimothyJones:

### PR Template

This fixes #1054 by widening the type so that the V3 matchers no longer need to be called with something that can be assigned to `AnyTemplate`. Since this is a widening of the type, it is not a breaking change.

I also removed all references to `AnyTemplate` and added tsdoc so that it is deprecated. It's still exported, but just not used anywhere. It can be removed in a future major release.

While I was there, I refactored the types out of matchers.ts into `types.ts`  for readability. These are now re-exported by `matchers.ts` so that any (inappropriate) deep links will still work, and anyone importing the type from `MatchersV3.Matcher` will still work.

I also added a TODO comment that in some future release we'll need to rename the `interfaces` that have the same names in the V2 and V3 matchers. This affects maintainers, as the current situation will result in confusing types, but I don't think it affects users.

Please don't squash merge this, as the commit messages contain the details for the release notes.

### DIFF
--- a/src/v3/matchers.ts
+++ b/src/v3/matchers.ts
@@ -1,39 +1,32 @@
 import { isNil, pickBy, times } from 'ramda';
 import RandExp from 'randexp';
+
+import {
+  AnyTemplate,
+  ArrayContainsMatcher,
+  DateTimeMatcher,
+  Matcher,
+  MaxLikeMatcher,
+  MinLikeMatcher,
+  ProviderStateInjectedValue,
+  V3RegexMatcher,
+} from './types';
+
 import { AnyJson, JsonMap } from '../common/jsonTypes';
 
-/**
- * Pact Matcher
- */
-export interface Matcher<T> {
-  'pact:matcher:type': string;
-  'pact:generator:type'?: string;
-  value?: T;
-}
-
-export function isMatcher(x: AnyTemplate): x is Matcher<AnyTemplate> {
+export function isMatcher(x: unknown): x is Matcher<unknown> {
   return (
     x != null &&
-    (x as Matcher<AnyTemplate>)['pact:matcher:type'] !== undefined &&
-    (x as Matcher<AnyTemplate>).value !== undefined
+    (x as Matcher<unknown>)['pact:matcher:type'] !== undefined &&
+    (x as Matcher<unknown>).value !== undefined
   );
 }
-
-export type AnyTemplate =
-  | AnyJson
-  | TemplateMap
-  | TemplateArray
-  | Matcher<unknown>;
-interface TemplateMap {
-  [key: string]: AnyJson | AnyTemplate;
-}
-type TemplateArray = Array<AnyTemplate>;
 
 /**
  * Value must match the given template
  * @param template Template to base the comparison on
  */
-export const like = <T extends AnyTemplate>(template: T): Matcher<T> => ({
+export const like = <T>(template: T): Matcher<T> => ({
   'pact:matcher:type': 'type',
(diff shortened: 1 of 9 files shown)


TimothyJones:
The build is failing because of unrelated reasons. The examples need to be updated (looks like mocha-pact needs updating too).

src/v3/matchers.ts
```diff
@@ -537,25 +495,38 @@ export const matcherValueOrString = (obj: unknown): string => {
...
  * example content.
  */
-export function reify(input: AnyTemplate): AnyJson {
+export function reify(input: unknown): AnyJson {
   if (isMatcher(input)) {
-    return reify(input.value as AnyTemplate);
+    return reify(input.value);
   }
 
-  if (Object.prototype.toString.call(input) === '[object Array]') {
```
mefellows, line 545: what on earth was going on here in the first place 😬

mefellows:
Hmm this is a new failure, I've not seen this one before but it seems like it exists in `master` so entropy has done its thing https://github.com/pact-foundation/pact-js/actions/runs/4190675102/jobs/7264320560#step:4:1477

I'm going to merge and sort out the other build after, and any issues with the types if they are discovered.

Thank you, this is much appreciated!

src/v3/matchers.ts
TimothyJones: Old school JS

TimothyJones:
That failure is because mocha-pact’s peer dependency is set to 9.x, so later versions of npm are pulling in the older one (and should also be failing on npm install- I guess npm ci doesn’t fail like install does, or something).

npm is getting better at peer dependencies, so we need to be more careful about them being correct

src/v3/matchers.ts
TimothyJones: I think Array.isArray is “reasonably” new

PR #1062, opened 2023-02-16
chore: Rename message's `StateHandlers` to `MessageStateHandlers`

### LINKED ISSUE #1057: Two StateHandler types defined
### Software versions

- **OS**: Mac OSX 13.1
- **Consumer Pact library**: Pact JS v10.4.1
- **Provider Pact library**: Pact JS v10.4.1
- **Node Version**: 18.12.1

### Issue Checklist

Please confirm the following:

- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures

[I raised this in the pact-js slack channel](https://pact-foundation.slack.com/archives/C9VBGLUM9/p1675340274779949?thread_ts=1674576330.071629&cid=C9VBGLUM9) and was asked to raise this issue.

### Expected behaviour

Creating StateHandlers [as per the documentation](https://docs.pact.io/implementation_guides/javascript/docs/provider#provider-state-setup-and-teardown) results in no Typing errors from TS.

### Actual behaviour

When attempting to use the setup/teardown properties on the provider state handlers Typescript complains that it doesn't match the type it's expecting. TS is finding two types with the same name and merging them (StateHandlers I believe). It gives the error:

```
(property) 'my state': {
    setup: () => Promise<void>;
    teardown: () => Promise<void>;
}
Type '{ setup: () => Promise<void>; teardown: () => Promise<void>; }' is not assignable to type 'StateHandler & ((state: string, params?: { [name: string]: string; } | undefined) => Promise<unknown>)'.
  Type '{ setup: () => Promise<void>; teardown: () => Promise<void>; }' is not assignable to type 'StateFuncWithSetup & ((state: string, params?: { [name: string]: string; } | undefined) => Promise<unknown>)'.
    Type '{ setup: () => Promise<void>; teardown: () => Promise<void>; }' is not assignable to type '(state: string, params?: { [name: string]: string; } | undefined) => Promise<unknown>'.
      Type '{ setup: () => Promise<void>; teardown: () => Promise<void>; }' provides no match for the signature '(state: string, params?: { [name: string]: string; } | undefined): Promise<unknown>'.ts(2322)
```

### Steps to reproduce

Create a state handler definition such as:

```
'Whatever your state name is': {
    setup: (parameters) => {
      // do your setup here
      // return a promise if you need to
    },
    teardown: (parameters) => {
      // do your teardown here
      // return a promise if you need to
    },
  },
```

TS will complain as above.

As a workaround, we've created a shim type. An alternative as mentioned by @TimothyJones in the slack thread is to define your own StateHandler type until this is fixed up.

### Relevant log files

N/A

TimothyJones:
In #882 the `StateHandlers` type was changed so that it was incompatible with the other `StateHandlers` type, resulting in conflicts when Typescript merged the two. This caused problems where it was impossible to correctly type `StateHandlers` (see  #1057). 

This PR is a breaking change that simply renames the message `StateHandlers` interface to `MessageStateHandlers`, and therefore fixes #1057.

Probably the better fix would be to introduce a breaking change that makes both types the same again - but I didn't want to get in to untangling what is supposed to happen.

This must not be squash merged, otherwise you won't get the breaking change in the changelog.

### DIFF
--- a/src/dsl/message.ts
+++ b/src/dsl/message.ts
@@ -85,7 +85,7 @@ export interface MessageProviders {
   [name: string]: MessageProvider;
 }
 
-export interface StateHandlers {
+export interface MessageStateHandlers {
   [name: string]: (
     state: string,
     params?: { [name: string]: string }
--- a/src/dsl/options.ts
+++ b/src/dsl/options.ts
@@ -4,7 +4,7 @@
  */
 import { VerifierOptions as PactCoreVerifierOptions } from '@pact-foundation/pact-core';
 import { PactfileWriteMode } from './mockService';
-import { MessageProviders, StateHandlers } from './message';
+import { MessageProviders, MessageStateHandlers } from './message';
 
 export type LogLevel = 'trace' | 'debug' | 'info' | 'warn' | 'error';
 
@@ -69,7 +69,7 @@ export interface MessageProviderOptions {
   messageProviders: MessageProviders;
 
   // Prepare any provider states
-  stateHandlers?: StateHandlers;
+  stateHandlers?: MessageStateHandlers;
 }
 
 type ExcludedPactNodeVerifierKeys = Exclude<


TimothyJones:
@mefellows  is there a reason this one is stalled? It affects everyone using typescript and states ... which these days might be everyone.

mefellows:
Thanks for the reminder, it just got lost in the pile - merging and releasing ASAP!

PR #1069, opened 2023-03-08
chore: Rename `uuid` matcher JSDocs param

jasonrowsell:
The current implementation of the `uuid` matcher function has a param name that does not match the JSDoc param. Although this doesn't cause any bugs, it may be good to update this for consistency.

Screenshot 2023-03-08 at 09 55 07

This PR renames the JSDoc param from `uuuid` to `id` to reflect the name of the param


TimothyJones (approved):
Good catch!

PR #1070, opened 2023-03-08
chore: pact-js-core 13.13.5

YOU54F:
Picks up https://github.com/pact-foundation/pact-js-core/pull/432 via 13.13.5 of `pact-js-core`

- side note, I imagine CI which is flakey which we should fix when we get time
- side side note, can we use dependabot to trigger these updates automatically to save a user manually doing it (if not I might script up quickly updating these values and running npm install in each of the respective folders)

### DIFF
--- a/examples/e2e/package-lock.json
+++ b/examples/e2e/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "devDependencies": {
         "@pact-foundation/pact": "10.1.2",
-        "@pact-foundation/pact-core": "^13.7.8",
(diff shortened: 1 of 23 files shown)


mefellows:
Will get this in today Yousaf, thanks folks.

> side side note, can we use dependabot to trigger these updates automatically to save a user manually doing it (if not I might script up quickly updating these values and running npm install in each of the respective folders)

Yes, that would be nice. I assume dependabot would discover this, but it might take a week or so (I'd need to review the configuration and defaults). I wonder if we could setup an automated build that is dependent on releasing `pact-core` that updates the library etc.

PR #1098, opened 2023-07-06
Upgrade pact js core prebuilds ci

YOU54F:
Includes and tests pact-core 13.15.0 which includes 
- https://github.com/pact-foundation/pact-js-core/pull/446

Requires for completeness
- https://github.com/pact-foundation/pact-js-core/pull/453

Some notes:-

1. some binary content issues as per https://github.com/pact-foundation/pact-js-core/issues/447
2. includes cirrus-ci to test arm64 on linux and macos
3. updates examples
  - skips mocha-pact and jest-pact examples until they are updated to work with the latest pact-js
  - updates an xml dep as it didn't have binaries for arm64 on the v3/e2e project
  - currently skips plugin tests on windows as verifier is failing, trying to isolate issue in this branch https://github.com/pact-foundation/pact-js/actions/runs/5479635684
  - reinstates macos tests on github actions, resolved one set of tests by increasing mocha timeout (had to update to .mocharc in the process and mocha-opts file is deprecated), removed clean up which was for the old ruby processes.


src/common/net.spec.ts
```diff
@@ -26,7 +26,9 @@ describe('Net', () => {
       });
     });
 
-  describe('#isPortAvailable', () => {
+  (process.platform === 'linux' && process.arch === 'arm64'
+    ? describe.skip
+    : describe)('#isPortAvailable', () => {
```
YOU54F, line 31: yak shave - maybe due to node slim image, need to try with full fat.

scripts/install-plugins.sh
```diff
@@ -13,6 +13,10 @@ function detect_osarch() {
             os='linux'
             arch='x86_64'
             ;;
+        'Linux aarch64')
+            os='linux'
+            arch='aarch64'
+            ;;
```
YOU54F, line 19: we can test on arm64 linux now

package.json
```diff
@@ -119,6 +119,7 @@
     "@babel/cli": "^7.18.6",
     "@babel/core": "^7.18.6",
     "@babel/preset-env": "^7.18.6",
+    "@cucumber/cucumber": "^9.1.2",
```
YOU54F, line 122: hmmm that is an accident. shall remove

examples/v3/provider-state-injected/package.json
```diff
@@ -11,9 +11,9 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@pact-foundation/pact": "10.1.2",
+    "@pact-foundation/pact": "11.0.2",
     "chai": "^4.3.6",
-    "jest": "^26.6.3"
```
YOU54F, line 16: jest 26 fails on the node:path lookup suggested and implemented  here

https://github.com/pact-foundation/pact-js-core/pull/446#discussion_r1193067918

updating to jest 29 resolves it, didn't test other versions in between

examples/v3/provider-state-injected/package-lock.json
YOU54F, line 1: didn't mean to remove

examples/v3/e2e/test/provider.spec.js
```diff
@@ -4,7 +4,7 @@ const chaiAsPromised = require('chai-as-promised');
 chai.use(chaiAsPromised);
 const { server, importData, animalRepository } = require('../provider.js');
 const path = require('path');
-const LOG_LEVEL = process.env.LOG_LEVEL || 'TRACE';
+const LOG_LEVEL = process.env.LOG_LEVEL || 'INFO';
```
YOU54F, line 7: it might be nicer to set info by default and set to trace on a workflow input in CI, so we get trace logs when we need them but the output is sane otherwise.

examples/v3/e2e/test/mocha.opts
YOU54F, line 1: deprecated, replaced by .mocharc.json

examples/v3/e2e/test/helper.ts
YOU54F, line 1: needed for ruby core, not needed for rust

.github/workflows/build-and-test.yml
```diff
@@ -27,50 +29,96 @@ jobs:
...
+      # - name: Upload dist folder
+      #   if: runner.os == 'Linux'
+      #   uses: actions/upload-artifact@v3
+      #   with:
+      #     path: dist
 
-  # Failures not reproducible locally on an M1 Mac
-  # build-and-test-macos:
-  #   runs-on: macos-latest
+  # examples:
```
YOU54F, line 77: I tried to see if I could package up the dist folder and distribute it to each of the examples in separate builds but I couldn't get what I wanted quickly enough and I was already several yaks deep

.github/workflows/build-and-test.yml
```diff
@@ -13,11 +13,13 @@ env:
...
 jobs:
-  build-and-test-ubuntu:
-    runs-on: ubuntu-latest
+  build-and-test:
+    runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x, 20.x]
```
YOU54F, line 21: we really should update the node engines both on here and pact-js-core

YOU54F:
note cirrus-ci failing on PR's

```
PactBroker::Client::Error - Command `git rev-parse --abbrev-ref HEAD` didn't return anything that could be identified as the current branch.
```

examples/v3/e2e/test/provider.spec.js
mefellows: agree, thanks.

src/common/net.spec.ts
mefellows: We're best keeping an eye on this, networking related bugs are a pain.

package.json
YOU54F: image

src/common/net.spec.ts
YOU54F: 110%

YOU54F:
> note cirrus-ci failing on PR's
> 
> ```
> PactBroker::Client::Error - Command `git rev-parse --abbrev-ref HEAD` didn't return anything that could be identified as the current branch.
> ```

Resolved by setting GIT_BRANCH and GIT_COMMIT via CIRRUS_* ENV vars.

info page for cirrus https://cirrus-ci.org/guide/writing-tasks/#environment-variables

best added into the pact-broker client, to capture these for Cirrus-CI users (of which we are for our arm64 builds :) )

.github/workflows/build-and-test.yml
YOU54F: updated in this pr with `fix!` commit prefix

examples/v3/e2e/test/helper.ts
mefellows: Yep! Can go.

.github/workflows/build-and-test.yml
mefellows: One herd at a time my friend! Let's pick this up in a subsequent piece of work.

src/common/net.spec.ts
mefellows: My guess is that it's more to do with what's running/available on the Cirrus CI environment, than anything arm64 specific. But it is a reminder of the unfortunate unreliability of that feature. I'm actually wondering if we need to deprecate/drop support for it, as in the current state it is not a straightforward x-platform feature.

mefellows: So I had a look at the failing test. It's failing, because we're expecting it _not_ to be able to find a free port on port `80`. But if it's running in a container (like Cirrus CI, I believe) then port `80` may well be allowed, and the test could pass. 

I'll see if we can ensure it's "unavailable" by other means.

YOU54F: oh yes I see now that on windows it sets `-1`, other platforms `80`. 

yes cirrus-ci in our current config is using a Dockerfile (or image) as the CI environment

YOU54F:
Going to get these changes merged in now and released, will get builds green.

- will update docs post release
- will raise tickets to track the yak shaves
  - provider plugin test on windows ci
  - is port available check in Docker environments

We are in a much better place than before, with testing support for additional arches, and finally covering macos in ci again

YOU54F:
note: as this change also bumps the node engines version, it constitutes a major change,

we could backport this change, without the node engines bump to v11.x, as this will release as v12.x

just a thought, otherwise troubleshooting notes will state v10/v11 of pact-js require the python build chain, v12 onwards doesn't. (it might be wiser leaving it as is, as we have only tested the prebuilds from node 16 onwards, so the engine bump imo makes sense (as is also present in pact-js-core, so a pact-js user on node <16 would probably see an error from pact-js-core)

mefellows:
Awesome! I agree with getting it in and bumping major version.

Issue #1031, opened 2022-12-22
Mismatch with header 'Accept'


danymarques:
### Software versions

- **OS**:  Mac OSX 13.0.1
- **Consumer Pact library**:  Pact JS v10.4.0
- **Node Version**: 14.21.1

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures
- [ ] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [x] For bonus points and virtual high fives, I have created a reproducible git repository (see below) to illustrate the problem

### Expected behavior

When giving multiple arguments to a header, the tests should work.

### Actual behavior

When we give multiple arguments to a header, eg: 
```
Accept: 'application/problem+json, application/json, text/plain, */*',
```
the tests fail with the following error:
```
1.0 Mismatch with header 'Accept': Expected header 'Accept' to have value 'application/problem+jsonapplication/jsontext/plain*/*' but was 'application/problem+json'
```
which means that the headers are not handled correctly.

### Steps to reproduce

* git clone git@github.com:danymarques/pact-error-demo.git
* npm ci
* npm run test

mefellows:
Thanks for this. I think it's a duplicate of #964. I've just pushed up a fix that should fix it, at least as far as we can in Pact JS.

You can now either specify headers as a comma delimited list (as you have in your repro) or explicitly as an `[]` of items, each of which can also have matching rules if desired.

I've noticed a small bug with the way it reports mismatches if the request header doesn't match, but you can see what's happening in the debug logs when enabled. I'll raise an upstream issue for that.

danymarques:
Hi @mefellows ,
I compiled your branch and tested it in the project and it works perfectly.
When are you planning to release it?
Cheers

GeorgeTailor:
@mefellows afair HTTP header does not have any constraints to be either a string or a comma-separated string or an array, for example, we use a JSON in a header with multiple properties, thus the object has commas. When it is sent as a string, some splitting magic happens and probably pact tries to treat broken parts of json as a separate array entries, while in reality it is a single string, but with commas.

mefellows:
[...]

PR #1099, opened 2023-07-08
fix: do not split headers by comma

mefellows:
This PR reverts the change added in #1031. Multi-valued headers are supported partially, as they must be specified using the array syntax. The open issues have connected and reproduced upstream issues now, but I think we should release this sooner rather than later.


YOU54F:
windows runners are hanging I believe on the v4 plugin provider verification. they aren't cancelling down, so locking up the orgs runners, so have cancelled the builds for now 👍🏾

PR #1102, opened 2023-07-13
Feat/status matcher

mefellows:
Not ready for merging, but FYI. 

To be resumed when https://github.com/pact-foundation/pact-reference/issues/296 is completed.

### DIFF
--- a/src/v3/types.ts
+++ b/src/v3/types.ts
@@ -68,6 +68,10 @@ export interface ProviderStateInjectedValue<T> extends Matcher<T> {
   expression: string;
 }
 
+export interface StatusCodeMatcher<T> extends Matcher<T> {
+  status: string | number[];
+}
+
 export interface RulesMatcher<T> extends Matcher<T> {
   rules: Matcher<T>[];
 }
@@ -147,7 +151,7 @@ export interface V3Request {
 }
 
 export interface V3Response {
-  status: number;
+  status: number | Matcher<number>;
   headers?: TemplateHeaders;
   body?: unknown;
   contentType?: string;
@@ -157,4 +161,31 @@ export interface V3MockServer {
   port: number;
   url: string;
   id: string;
+}
+
+export enum HTTPResponseStatusClass {
+  // Informational responses (100–199)
+  Information = 'information',
+  // Successful responses (200–299)
+  Success = 'success',
+  // Redirects (300–399)
+  Redirect = 'redirect',
+  // Client errors (400–499)
+  ClientError = 'clientError',
+  // Server errors (500–599)
+  ServerError = 'serverError',
+  // Informational responses (100–199)
+  '1XX' = 'information',
+  // Successful responses (200–299)
+  '2XX' = 'success',
+  // Redirects (300–399)
+  '3XX' = 'redirect',
+  // Client errors (400–499)
+  '4XX' = 'clientError',
+  // Server errors (500–599)
+  '5XX' = 'serverError',
+  // Non-error response(< 400)
+  NonError = 'nonError',
+  // Any error response (>= 400)
+  Error = 'error',
 }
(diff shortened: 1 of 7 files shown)


src/v3/types.ts
```diff
@@ -154,3 +162,20 @@ export interface V3MockServer {
...
+  // Redirects (300–399)
+  Redirect = 'redirect',
+  // Client errors (400–499)
+  ClientError = 'clientError',
+  // Server errors (500–599)
+  ServerError = 'serverError',
+  // Non-error response(< 400)
+  NonError = 'nonError',
+  // Any error response (>= 400)
+  Error = 'error',
```
TimothyJones, line 180: I think these would be better named `2XX`, `4XX`, etc.

src/v4/http/index.ts
```diff
@@ -116,8 +116,11 @@ export class InteractionwithRequest implements V4InteractionwithRequest {
     private opts: PactV4Options
   ) {}
 
-  willRespondWith(status: number, builder?: V4ResponseBuilderFunc) {
-    this.interaction.withStatus(status);
+  willRespondWith(
+    status: number | Matcher<number>,
+    builder?: V4ResponseBuilderFunc
```
TimothyJones, line 121: Unrelated, but is the builder pattern the right thing to have in this API? It's not very idomatic.

src/v4/http/index.ts
```diff
@@ -296,7 +299,7 @@ export class InteractionWithPluginRequest
     status: number,
     builder?: V4PluginResponseBuilderFunc
   ): V4InteractionWithPluginResponse {
-    this.interaction.withStatus(status);
+    this.interaction.withStatus(reify(status) as number);
```
TimothyJones, line 302: I suggest removing the `as number`. You could do this by adding a type parameter to `reify`.

src/v4/http/types.ts
```diff
@@ -71,7 +72,7 @@ export interface V4InteractionWithCompleteRequest {
 
 export interface V4InteractionwithRequest {
   willRespondWith(
-    status: number,
+    status: number | Matcher<number>,
```
TimothyJones, line 75: Is this right? `Matcher<number>` isn't necessarily the same as `StatusCodeMatcher<T>`.

Also, if these aren't always http requests, then I'm not sure that `number` is right

src/v4/http/types.ts
```diff
@@ -137,7 +138,7 @@ export interface V4InteractionWithPlugin {
 
 export interface V4InteractionWithPluginRequest {
```
TimothyJones, line 139: Unrelated, but does this need to be a separate interface?

TimothyJones (commented):
A status matcher would be great! I left some musings inline.

src/v3/types.ts
mefellows: I had considered that also, I'm glad you mentioned it.

I thought that might be too casual, so I stuck with the definitions in the rust core. Presumably, the first 5 at least were named after the actual classes (e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status).

src/v4/http/index.ts
mefellows: I definitely ummed and ahh'd about this. It could probably be a well defined type (or possibly set of types). The issue with using TypeScript types to represent different possible inputs, is that people still use JS and eventually put in garbage, so you need validation on top of it. 

The builder pattern doesn't allow it, and saves all of this additional validation code, whilst also still giving you a typed interface. 

But I'll revisit before it's no longer behind a flag. Thanks.

src/v4/http/types.ts
mefellows: It can't be a non HTTP request, the type-state designed DSL/interface won't allow that.

mefellows: ...but your other point is correct, thank you.

src/v4/http/types.ts
mefellows: It does, as the builder func is different and you must provide certain details prior to this being an allowed state. See the type-state pattern, which I think resolves so many problems with the current DSL (and with the more complicated options allowed in V4, such as plugins).

Issue #1077, opened 2023-03-23
eachKeyLike() creates wrong matchingRules 


eburi:
### Software versions

_Please provide at least OS and version of pact-js_

- **OS**:  Mac OSX 13.1 (tested on Windows 11 as well)
- **Consumer Pact library**: 11.0.2 (same with latest v10.x)
- **Provider Pact library**: n/a
- **Node Version**: v18.15.0

```
    "pactRust": {
      "ffi": "0.4.0",
      "models": "1.0.4"
    },

``` 

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures
- [x] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [x] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour

Given this interaction:
```
  interaction('pact integration', ({ provider, execute }) => {
    beforeEach(() =>
      provider
        .given('A pet 1845563262948980200 exists')
        .uponReceiving('A get request to get a pet 1845563262948980200')
        .withRequest({
          method: 'GET',
          path: '/v2/pet/1845563262948980200',
          headers: { api_key: '[]' },
        })
        .willRespondWith({
          status: 200,
          // Hypothetical response from an API
          // body: {
          //   options: {
          //     'DEFAULT': {
          //       value: MatchersV3.integer(12),
          //       selected: MatchersV3.boolean(false)
          //     },
          //     'SUPER': {
          //       value: MatchersV3.integer(120),
          //       selected: MatchersV3.boolean(true)
          //     },
          //     'LUXERY': {
          //       value: MatchersV3.integer(12000),
          //       selected: MatchersV3.boolean(false)
          //     }
          //   }
          // }

          // Pact for this API
            body: {
              options: MatchersV3.eachKeyLike('SUPER', {
                value: MatchersV3.integer(120),
                selected: MatchersV3.boolean(true),
              }),
              optionArray: MatchersV3.eachLike({
                value: MatchersV3.integer(120),
                selected: MatchersV3.boolean(true),
              }),
            },
        })
    );
```

I would expect matchingRules in the resulting pact-file for the body that look like this:
```
[...]

PR #1103, opened 2023-07-13
Add support for `eachKey` and `eachValue` matchers

### LINKED ISSUE #952: `eachKeyLike` matcher isn't eachKeyLike, it is eachValueLike (also rust core version needs a bump)
### Software versions

- **OS**: All
- **Consumer Pact library**: 10.1.4
- **Provider Pact library**:N/A
- **Node Version**: N/A
- 
### Issue Checklist

Please confirm the following:

- [X] I have upgraded to the latest
- [X] I have the read the FAQs in the Readme
- [X] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures

### Expected behaviour

As per https://github.com/pact-foundation/pact-reference/issues/216 , `eachLike` doesn't match correctly.  This object:

```
{
  "john-doe1": {
    "brown-fox": {
      "jumps": "over",
      "the": "lazy dog"
    }
  },
  "john-doe2": {
    "brown-fox2": {
      "jumps": "over",
      "the": "lazy dog"
    }
  }
}
```
should be matched by

```
Matchers.eachKeyLike(
   "some string", 
   Matchers.eachKeyLike(
      "some-string", 
      Matchers.eachKeyLike(
         "some-string",
         Matchers.string())))
```

### Actual behaviour

It is not matched - the match only applies at the first level.

### Steps to reproduce

As above

### Extra information

To get the fix, we'll need to bump the rust core to at least 0.3.12.

Also, as @uglyog pointed out:

> BTW, the eachKeyLike function is incorrectly named. It should be something like eachValueLike.

Probably the right fix for this is:

* Rename the existing `eachKeyLike` to `eachValueLike`
* Add the `eachKeyLike` matcher (this is a breaking change, since the signature for `eachKeyLike` would describe something different to what it currently does / and currently is documented to do)
* Update the documentation to match
* For bonus points, I think there's also an `eachKeyValueLike` matcher which could be added

mefellows:
* Deprecates the `eachKeyLike` matcher, which is broken (#1077, #662)
* Adds `eachKeyMatches` to match on keys (the `like` part is dropped, because it's not a `like` matcher, it can accept multiple)
* Adds `eachValueMatches` to match on values (counterpart to `eachKeyMatches`)

Fixes #952


TimothyJones:
> Deprecates the eachKeyLike matcher, which is broken 

This deprecation doesn't appear to be mentioned in the release notes

mefellows:
Thanks, good spot. I can fix that manually.

PR #1111, opened 2023-08-28
fix: MessageConsumerPact merge pact file correctly

matdodgson:
### PR Template

This PR fixes the merge parameter for the call to writePactFile() in MessageConsumerPact. The change matches the code for the same call from the Http Pact.


mefellows:
Thanks!

mefellows:
Yeah, I can't see why your changes would have broken. I'll need to dig into what's happening, but happy to merge your changes.

PR #1124, opened 2023-10-17
Gql examples

YOU54F:
Updates to the GraphQL example to use lots of the latest deps, including some deprecation fixups

Tested with node 18 / macos arm64


YOU54F:
I'll yak shave the failing CI examples this week, there are some failing on master as well

PR #1165, opened 2024-01-18
fix: update pact core dependency

miguelci:

### PR Template

Update the pact-core dependency to their last version.
https://github.com/pact-foundation/pact-js-core/releases/tag/v14.0.6

There was a fix on one of its updates on the `fromProviderState` function that might fix an issue mentioned in https://pact-foundation.slack.com/archives/C9VBGLUM9/p1705297213981799?thread_ts=1704896397.421819&cid=C9VBGLUM9

### DIFF
--- a/examples/e2e/package-lock.json
+++ b/examples/e2e/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "devDependencies": {
         "@pact-foundation/pact": "11.0.2",
-        "@pact-foundation/pact-core": "^14.0.1",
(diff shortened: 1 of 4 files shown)


mefellows:
This PR isn't really needed, as the dependency in the node allows patch version updates. If you need the latest, the simplest thing to do is an `npm update @pact-foundation/pact` in the project and you should get the latest.

> can someone confirm if the fails are related to some issue in the server or to the change?

hmm that is a curious failure, I've actually never seen it before. I'll see if I can reproduce locally.

PR #1168, opened 2024-01-22
fix: support multiple states with same name

### LINKED ISSUE #848: [V3] provider states with the same name are not exported to pact file
### Software versions

- **OS**:  Ubuntu 21.04
- **Consumer Pact library**: Pact JS v10.0.0-beta.59
- **Provider Pact library**:  Pact JS v10.0.0-beta.59
- **Node Version**: v16.0.0

### Issue Checklist

Please confirm the following:

- [ ] I have upgraded to the latest
- [ ] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/#test-fails-when-it-should-pass) the section on intermittent test failures

### Expected behaviour

all provider states are saved to JSON

This would be useful when the provider test needs more than one set of testdata of the same type to be useful. E.g. create multiple files and list them, create multiple users, etc.

### Actual behaviour

only one provider state is saved to JSON

### Steps to reproduce

create multiple provider states with the same name
My use-case is to create multiple files. So I call multiple times 
`provider.given('file exists', { username, password, fileName: resource })`

but in the JSON I only have one providerstate with that name
```
      "providerStates": [
        {
          "name": "the user is recreated",
          "params":...
        },
        {
          "name": "file exists",
          "params": {
            "fileName": "testFolder/文件.txt",
            "password": "test123",
            "username": "test123"
          }
        },
        {
          "name": "resource is shared",
          "params": ...
        }
      ]
```

Making the names of provider states individual makes them appear in the JSON, but that would mean I would also have to have an individual state definition in the provider tests (stateHandlers)

mefellows:
Fixes #848 using the new `ffi_given_with_params` core function.

Allows different parameters for the same state name to be reused, resulting in multiple provider states with the same description, but different parameters.

### DIFF
--- a/examples/e2e/package-lock.json
+++ b/examples/e2e/package-lock.json
@@ -9,8 +9,7 @@
       "version": "1.0.0",
       "license": "MIT",
       "devDependencies": {
-        "@pact-foundation/pact": "11.0.2",
-        "@pact-foundation/pact-core": "^14.0.1",
+        "@pact-foundation/pact": "file:../../dist",
         "absolute-version": "1.0.1",
         "body-parser": "^1.20.0",
         "chai": "^4.3.6",
@@ -26,13 +25,13 @@
     },
     "../../dist": {
       "name": "@pact-foundation/pact",
-      "version": "11.0.2",
+      "version": "12.1.2",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@pact-foundation/pact-core": "^13.15.0",
+        "@pact-foundation/pact-core": "^14.1.1",
         "@types/express": "^4.17.11",
-        "axios": "^0.27.2",
+        "axios": "^1.6.1",
         "body-parser": "^1.20.0",
         "cli-color": "^2.0.1",
         "express": "^4.18.1",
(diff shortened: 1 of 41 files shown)


mefellows:
Crap, the `pactffi_given_with_params` accepts 3 args not 4! I need to do the whole release shenanigans again :P

YOU54F:
doh! I'll get on that today

YOU54F:
Pact-js-core release sorted, and changes pulled into this branch from master, so should be good to go.

Would be good to include a test case that covers this, so we can say its definitely boxed off, maybe we can dust off a repro from the original issue

mefellows:
Thanks, yes probably need a smoke test suite for the FFI functions in the core (to prevent that bad mapping). Ideally, they are auto generated - just needs some solid spiking time to generate a workable concept.

In Pact JS, the function will be used in lots of places so will get the coverage needed (I think).

As for the failing tests, strange they work on some OS/Arch combinations. Thanks for looking (and raising the upstream in pact reference).

YOU54F:
windows currently skip the v4 matt plugin test as they were failing in ci (fine locally)

macos/linux covers them which is why are they are red, and the windows are green.

YOU54F (approved):

mefellows:
Thank you sir!

PR #1172, opened 2024-01-25
ci: remove cirrus whilst over limits

YOU54F:
Currently over limits on Cirrus CI, which is causing GH status checks to fail.

Whilst we haven't sorted out a permanent solution, I've parked the current cirrus config on the `cirrus_ci` branch, and propose deleting it from the main branch in the interim


YOU54F:
Will be switching to GHA M1 runners shortly

https://github.com/pact-foundation/pact-js-core/pull/478

Issue #1169, opened 2024-01-23
High Prototype Pollution risk caused by lodash.omitby/4.6.0 scanned by BlackDuck for @pact-foundation/pact@^12.1.2


Rufei77:
_Thank you for reporting a bug! We appreciate it very much. Issues are a big input into the priorities for Pact-JS development_

_All italic text in this template is safe to remove before submitting_

_Thanks again!_

### Software versions

_Please provide at least OS and version of pact-js_

- **OS**: Mac OS
- **Consumer Pact library**: @pact-foundation/pact@^12.1.2
- **Provider Pact library**: @pact-foundation/pact@^12.1.2
- **Node Version**: v 18.xx

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures
- [ ] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [ ] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour

No vulnerabilities reported :)

### Actual behaviour

Blackduck scanner report a HIGH severity alert ([CVE-2019-10744](https://nvd.nist.gov/vuln/detail/CVE-2019-10744) for a dependency (lodash.omitby/4.6.0) used by pact.

mefellows:
Thanks, are you interested in fixing this? Either by upgrading the dependency or replacing `omitBy` with another function?

FWIW you should consider and discuss whether or not a developer dependency is really exploitable and a HIGH severity risk (I bet it isn't).

yukun-han:
Hi @mefellows , I am one of @Rufei77 's colleagues and I'm here to help her raise a PR to fix this issue. The PR #1175 is already linked here. Please take your time to have a look and feel free to give feedbacks.

Further to discuss, `lodash` is not actively maintained now. As times going, more and more security risks would possibly be reported by vulnerability scanning tools like Snyk, BlackDuck and no one would go to take care of them! In my perspective, it is worthwhile to retire all `lodash` dependencies and replace with alternatives. I noticed that `ramda` is also listed in dependencies. It is a good choice.

mefellows:
I think this may be closed now that the other item has been merged and released - thanks for the PR! Closing.

PR #1175, opened 2024-02-03
fix: retire lodash omitby to fix vulnerability

yukun-han:
# Checklist

# Summary

Mentioned in issue #1169, `lodash.omitby` will be reported security risk by BlackDuck. This PR retired `lodash.omitby` dependency and replaced it with `ramda` to avoid vulnerability.

`lodash.omitby` is used to filter out null and undefined value from a object. `ramda` does not provide same name `omitBy` function but this functionality could be achieved by `ramda.reject` referring to [link](https://github.com/ramda/ramda/issues/2084).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,6 @@
         "lodash.isnil": "4.0.0",
         "lodash.isundefined": "3.0.1",
         "lodash.omit": "^4.5.0",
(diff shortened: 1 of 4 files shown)


mefellows:
Thank for you submitting this, if the build passes I'll bring this in.

Issue #1188, opened 2024-03-13
Bug: Corporate Proxy not supported during verification


mefellows:
### Discussed in https://github.com/pact-foundation/pact-js/discussions/1187

<div type='discussions-op-text'>

<sup>Originally posted by **zac-poe** March  7, 2024</sup>
I am in the process of updating to the latest version of pact-js and I am encountering issues configuring a web proxy for use with the Verifier.

My approach with pact 9.17 (and pact-node 10.17) was to use global-agent to leverage our https proxy and this worked successfully when my providerBaseUrl for the pact Verifier would otherwise be blocked by our web proxy.

When updating to pact 12.3.0, I noticed several issues with the above approach. Initially I had to add NO_PROXY configuration for 127.0.0.1. Subsequently, I saw runtime issues with the pact parseBody.js behavior as the proxy request was already in a finished/headersSent state when this event hook was attempting to update the request. However, even after manually modifying this behavior I was observing the global agent proxy on https was still not actually being leveraged and my provider requests were not going through the web proxy.

Similary, setting environment variables for HTTPS_PROXY or HTTP_PROXY does not seem to impact the underlying pact internal proxy forwarding from local to the provider URL.

What is the proper way to configure use of a corporate proxy with the latest Pact Verifier? Is this a feature that would need to be requested as an enhancement (e.g. adding an optional arg for an http Agent to the Verifier)?</div>

PR #1199, opened 2024-03-18
feat(corporate proxy #1188): support in verifier

zac-poe-vg:
Add native support of HTTPS_PROXY and HTTP_PROXY for verifier for #1188 

This change adds two pieces of behavior:
- The verifier's 2 step parse body behavior now uses a proxy request buffer (instead of the 'proxyReq' rewrite behavior) to ensure the request stream is accessible for intermediary agent behavior
- The verifier natively supports HTTP_PROXY/HTTPS_PROXY config and assigns a proxy http Agent where those values are present

This results in out-of-the-box support for corporate proxies in the js verifier

### DIFF
--- a/src/dsl/verifier/proxy/parseBody.ts
+++ b/src/dsl/verifier/proxy/parseBody.ts
@@ -4,24 +4,18 @@ interface ReqBodyExtended extends http.IncomingMessage {
   body?: Buffer | Record<string, unknown>;
 }
 
-export const parseBody = (
-  proxyReq: http.ClientRequest,
-  req: ReqBodyExtended
-): void => {
-  if (!req.body || !Object.keys(req.body).length) {
-    return;
-  }
+export const parseBody = (req: ReqBodyExtended): Buffer => {
+  let bodyData = Buffer.alloc(0);
 
-  let bodyData;
+  if (!req.body || !Object.keys(req.body).length) {
+    return bodyData;
+  }
 
   if (Buffer.isBuffer(req.body)) {
     bodyData = req.body;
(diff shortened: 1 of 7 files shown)


src/dsl/verifier/proxy/parseBody.ts
```diff
@@ -4,24 +4,18 @@ interface ReqBodyExtended extends http.IncomingMessage {
...
-
   if (Buffer.isBuffer(req.body)) {
     bodyData = req.body;
   } else if (typeof req.body === 'object') {
-    bodyData = JSON.stringify(req.body);
+    bodyData = Buffer.from(JSON.stringify(req.body));
   }
 
-  if (bodyData) {
-    proxyReq.setHeader('Content-Length', Buffer.byteLength(bodyData));
```
zac-poe-vg, line 24: this is one piece of behavior that did go away - I didn't see any tests around content length specifically, so I thought this might have just been a best practice with the rewrite that was used here

if this creates a gap, let me know and I can see about doing something similar with the proxy request options

mefellows: Thanks. I believe this was required given the way the previous body was parsed and then re-written into the subsequent request. Given that you're now using the `buffer` option on the `HttpProxy` itself, perhaps it's no longer needed. 

The main behaviour we'll need to check is if we can intercept and modify the incoming body using the hooks/filters. I'll pull this in locally and do some testing, I don't think it's an area with great test coverage if I'm honest.

mefellows (commented):
This looks great. i'm going to test locally with a few scenarios and if they work, will bring this in asap!

PR #1212, opened 2024-04-23
fix: update to latest pact-core 14.3.4

YOU54F:
Updates pact-core to 14.3.4 and updates package.json refs so all the examples are tested against the latest versions.

Note:- we have a caret in the package.json so I don't believe we need a new published version of pact-js.

We should add a macos-14 runner to test against arm64, as I get an error locally on an m1 mbp running the v4 matchers test

```console
[14:48:11.902] ERROR (75855): pact@12.4.0: Test failed for the following reasons:

  Mock server failed with the following mismatches:

        0) The following request was incorrect: 

                GET /eachKeyMatches
            
                         1.0 Expected query parameter 'from' but was missing
                         1.1 Mismatch with header 'Accept': Expected header 'Accept' at index 1 to have value '' but was 'text/plain'
                         1.2 Mismatch with header 'Accept': Expected header 'Accept' at index 2 to have value '' but was '*/*'
      1) returns the shape of object given to the matcher


  0 passing (47ms)
  1 failing

  1) V4 Matchers
       eachKeyMatches
         returns the shape of object given to the matcher:
     AxiosError: Request failed with status code 500
      at settle (/Users/yousaf.nabi/dev/pact-foundation/pact-js/examples/v4/matchers/node_modules/axios/lib/core/settle.js:19:12)
      at IncomingMessage.handleStreamEnd (/Users/yousaf.nabi/dev/pact-foundation/pact-js/examples/v4/matchers/node_modules/axios/lib/adapters/http.js:589:11)
      at IncomingMessage.emit (node:events:526:35)
      at IncomingMessage.emit (node:domain:488:12)
      at endReadableNT (node:internal/streams/readable:1589:12)
      at processTicksAndRejections (node:internal/process/task_queues:82:21)
```


mefellows:
Thanks! As noted, strictly we don't need it, but it's nice to bump these every now and then.

mefellows:
Interesting, just got that failure here during the release step: https://github.com/pact-foundation/pact-js/actions/runs/8842703164/job/24281763224#step:4:6248

This runs on `ubuntu-latest` node 18, which previously passed in the build step 🤔 .

mefellows:
Strange it worked at all, I've fixed the test - it wasn't setting the headers and it wasn't setting the query param.

mefellows:
OK I see, the `build-and-test.yml` has each example as a separate step, and is unsurprisingly missing this example:
Screenshot 2024-04-26 at 1 51 27 PM

The release step runs a different script, which automagically picks up all examples - hence why it only fails during the release stage!

PR #1242, opened 2024-10-05
fix: support async afterEach hook

### LINKED ISSUE #1241: afterEach does not support async operation
### Software versions

- **OS**: Mac OS Sonoma 14.7
- **Consumer Pact library**: Pact JS v13.1.13
- **Provider Pact library**: Pact JS v13.1.13
- **Node Version**: 22.9.0

### Issue Checklist

Please confirm the following:

- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures

### Expected behaviour

When defining an async function for `afterEach`, it should pause further execution of the verification tests until that Promise has resolved.

We require async behaviour for clearing up state in `afterEach` as part of our verification tests.

### Actual behaviour

Test execution is not paused, and Promise is not handled cleanly.

### Steps to reproduce

* Install the [sample project](https://github.com/mefellows/pact-js-repro-template) as a basis for this test
* Upgrade project to latest pact-js version (`npm i -DE @pact-foundation/pact@latest`)
* Modify `provider.spec.js` to configure the provider verifier as so:
```js
const pact = new pact_1.Verifier({
  pactUrls: ["./pacts/myconsumer-myprovider.json"],
  providerBaseUrl: "http://127.0.0.1:8080",
  afterEach: async () => {
    return new Promise((resolve) => setTimeout(resolve, 2000))
  },
});
```
* Run provider test (`./node_modules/.bin/jest provider.spec.js`)

Given this `afterEach` hook, I would expect test execution to pause for ~2 seconds, but it doesn't and jest does not exit cleanly, giving the warning:

```
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

Note the timestamps either side of the `executing 'afterEach' hook` log indicate execution on this thread is no paused ...

```
2024-10-04T16:15:22.434564Z DEBUG ThreadId(02) verify_interaction{interaction="a request for dynamic keys"}: hyper_util::client::legacy::pool: reuse idle connection for ("http", 127.0.0.1:60185)
[17:15:22.434] DEBUG (53817): pact@13.1.3: incoming request: {"body":{},"headers":{"accept":"*/*","accept-encoding":"gzip, deflate","host":"127.0.0.1:60185"},"method":"GET","path":"/"}
[17:15:22.434] DEBUG (53817): pact@13.1.3: Proxying GET: /
[17:15:22.436] DEBUG (53817): pact@13.1.3: executing 'afterEach' hook
[17:15:22.438] DEBUG (53817): pact@13.1.3: outgoing response: {"body":"{\"foo\":\"1\"}","headers":{"x-powered-by":"Express","content-type":"application/json; charset=utf-8","content-length":"11","etag":"W/\"b-Pvi0Q5/yHZyhiaQpID3i4+RNmB4\"","date":"Fri, 04 Oct 2024 16:15:22 GMT","connection":"close"},"status":200}
```

Note that `beforeEach` _does_ appear to support async operation. Here are the equivalent logs when replacing `afterEach` with `beforeEach` in the above snippet, which show a ~2 second gap between logs:

```
2024-10-04T16:23:18.834894Z DEBUG ThreadId(02) verify_interaction{interaction="a request for dynamic keys"}: hyper_util::client::legacy::connect::http: connected to 127.0.0.1:60460
[17:23:18.845] DEBUG (58216): pact@13.1.3: executing 'beforeEach' hook
[17:23:20.849] DEBUG (58216): pact@13.1.3: incoming request: {"body":{"action":"setup","params":{},"state":""},"headers":{"content-type":"application/json","accept":"*/*","accept-encoding":"gzip, deflate","host":"127.0.0.1:60460","content-length":"41"},"method":"POST","path":"/_pactSetup"}
```

lhokktyn:
Currently a asynchronous afterEach hook will execute concurrently with the underlying express request lifecycle.

This fix allows async hooks to complete before allowing the express request to continue through its middleware stack.

Resolves #1241


src/dsl/verifier/proxy/hooks.ts
```diff
@@ -8,21 +8,21 @@ export const registerBeforeHook = (
...
+      try {
+        await config.beforeEach();
+        next();
+      } catch (e) {
+        logger.error(`error executing 'beforeEach' hook: ${e.message}`);
+        logger.debug(`Stack trace was: ${e.stack}`);
+        next(new Error(`error executing 'beforeEach' hook: ${e.message}`));
       }
+    } else {
+      next();
```
TimothyJones, line 24: Good catch, this one was wrong too, but in a different way 😅 .

src/dsl/verifier/proxy/hooks.ts
```diff
@@ -31,19 +31,17 @@ export const registerAfterHook = (
   config: ProxyOptions,
   stateSetupPath: string
 ): void => {
+  logger.trace("registered 'afterEach' hook");
```
TimothyJones, line 34: This is now slightly misleading- it always registers the hook, but the hook is only connected to something if `config.afterEach` is set.

TimothyJones (approved):
Nice! Thanks for this. I'm not a maintainer any more, but it looks right to me.

lhokktyn:
@mefellows just a heads up that I'm lining up another PR that builds on top of this PR, which contains some further refactoring around the hooks stuff - mainly with the aim of ensuring hooks are only run once during each interaction's setup/teardown phase respectively.

I've got it marked as a breaking change at the moment - hence wanting to raise as a separate PR - but I'm not sure it is actually a break to be honest so a little discussion around that would be great once this PR is merged if you don't mind.

YOU54F (approved):

YOU54F:
> @mefellows just a heads up that I'm lining up another PR that builds on top of this PR, which contains some further refactoring around the hooks stuff - mainly with the aim of ensuring hooks are only run once during each interaction's setup/teardown phase respectively.
> 
> I've got it marked as a breaking change at the moment - hence wanting to raise as a separate PR - but I'm not sure it is actually a break to be honest so a little discussion around that would be great once this PR is merged if you don't mind.

That makes sense, looking forward to seeing the change.

Thanks for the repro and fix. I'll get this merged and released today

TimothyJones:
FWIW, I took a look at your branch and it looks more like a bugfix to me.

> Current behaviour will run the beforeEach and afterEach hooks
> multiple times if there are several provider states defined in
> an interaction. This change will ensure each of those hooks is
> run only once, regardless of how many provider states are
> defined in an interaction.

Yes, technically this is a break (as are most bug fixes, as they change behaviour), but the test I would usually apply is:

* Would anyone reasonably expect the old behaviour? (I don't think so)
* Would anyone reasonably rely on the old behaviour? (I don't think so)
* Was it already documented to work the old way? (no, it was documented to work the fixed way [here](https://github.com/pact-foundation/pact-js/blob/master/docs/provider.md#provider-state-setup-and-teardown))

I think it's still a bit surprising that the afterEach runs at the start of the state teardowns rather than after all of them, but it's hard to do that with Pact-JS's proxy.

Really, the afterEach should be first class options run by the pact core, so all languages get them, and the core can reason about the interactions with the teardowns too. This was one of the design philosophies I took when designing [ContractCase](https://case.contract-testing.io/docs/intro)  - which won't work for you here, as it doesn't have afterEach / beforeEach methods at all yet.

TimothyJones:
> I think it's still a bit surprising that the afterEach runs at the start of the state teardowns rather than after all of them, but it's hard to do that with Pact-JS's proxy.

You might be able to count the state setup calls, and rely on the same number of teardowns, maybe.

lhokktyn:
Opened https://github.com/pact-foundation/pact-js/pull/1243 which takes your comments into account.

mefellows:
> @mefellows just a heads up that I'm lining up another PR that builds on top of this PR, which contains some further refactoring around the hooks stuff - mainly with the aim of ensuring hooks are only run once during each interaction's setup/teardown phase respectively.

> I've got it marked as a breaking change at the moment - hence wanting to raise as a separate PR - but I'm not sure it is actually a break to be honest so a little discussion around that would be great once this PR is merged if you don't mind.

Thanks for this, I agree with the comments above - I wouldn't consider this a breaking change. It's incidentally working this way currently, and for most people this is unexpected behaviour.

Thanks for merging Yousaf, I wanted to test myself before merging as middleware can be hard to reason about by just looking at the code and it's not an area that is well covered by tests. I was concerned it 

> Really, the afterEach should be first class options run by the pact core, so all languages get them, and the core can reason about the interactions with the teardowns too

I agree @TimothyJones and created this issue to track that: https://github.com/pact-foundation/pact-reference/issues/468 (I thought I had previously done this, but it seems like it was a figment of my imagination).

PR #1243, opened 2024-10-08
fix: allow hooks to run only once per interaction

lhokktyn:
Current behaviour will run the `beforeEach` and `afterEach` hooks multiple times if there are several provider states defined in an interaction. This is due to the way the hook are run on every call to `/_pactSetup` via the proxy.

This change will ensure each of those hooks is run only once per interaction, regardless of how many provider states are defined in that interaction.

---

Our primary use case in applying this fix was to be able to support multiple `given()` statements in our consumer tests without having them each trigger the `beforeEach` hook (which we want to use for resetting state), eg

```js
await provider()
  .addInteraction()
  .given("has a pet with status", "sold")
  .given("has a pet with status", "pending")
  .uponReceiving("a request for pets with multiple statuses")
  ...
```

Which produces the following in the Pact JSON:

```json
{
  "description": "a request for pets with multiple statuses",
  "pending": false,
  "providerStates": [
    {
      "name": "has a pet with status",
      "params": {
        "value": "sold"
      }
    },
    {
      "name": "has a pet with status",
      "params": {
        "value": "pending"
      }
    }
  ],
  ...
}
```

Which drives multiple calls to the `/_pactSetup` endpoint during Provider verification, one for each of the `providerStates`, and thus multiple calls to `beforeEach`.

This fix ensures the hooks are only run on the first `setup` call and on the last `teardown` call.

### DIFF
--- a/src/dsl/verifier/proxy/hooks.spec.ts
+++ b/src/dsl/verifier/proxy/hooks.spec.ts
@@ -1,0 +1,121 @@
+import { expect } from 'chai';
+import { stub } from 'sinon';
+import { RequestHandler } from 'express';
+
+import {
+  registerHookStateTracking,
+  registerBeforeHook,
+  registerAfterHook,
+  HooksState,
+} from './hooks';
+
+// This mimics the proxy setup (src/dsl/verifier/proxy/proxy.ts), whereby the
+// state handling middleware is run regardless of whether a hook is registered
+// or not.
+const doRequest = async (
+  action: string,
+  hooksState: HooksState,
+  hookHandler?: RequestHandler
+) => {
+  const hooksStateHandler = registerHookStateTracking(hooksState);
+  const hookRequestHandler = hookHandler || ((req, res, next) => next());
+
+  const request: any = {
+    body: {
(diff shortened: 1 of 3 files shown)


lhokktyn:
@mefellows I'm not sure why all those checks have failed on the last push, as only the git commit message was altered. Maybe a temporary issue with the test broker?

lhokktyn:
Hi @mefellows . Do you think this may be a good candidate for merging? Let me know if you'd like anything revisiting. Cheers

mefellows:
Hi @lhokktyn - apologies for my tardiness, I'd like to test this out locally before merging. At a glance it looks good and I'm very appreciative of the tests also. I'll aim to get this out early next week.

lhokktyn:
Hey @mefellows no worries, appreciate your time :)

mefellows:
Sorry for the delay here @lhokktyn - this change has been released into the latest version v13.1.5.

lhokktyn:
Brilliant! Thanks for the heads up @mefellows 👍

PR #1247, opened 2024-11-14
chore: correct package homepage link in npmjs.com

longdoan7421:
## Summary
The homepage link for the `@pact-foundation/pact` package is currently incorrect. It points to https://docs.pact.io/implementation_guides/javascript/, which results in a "Page Not Found" error.

This pull request updates the homepage link to the correct URL: https://docs.pact.io/implementation_guides/javascript/readme.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
   "bugs": {
     "url": "https://github.com/pact-foundation/pact-js/issues"
   },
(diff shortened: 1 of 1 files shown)


YOU54F:
Thanks for letting us know and your contribution, I am going to solve this by adding a redirect to the docs site, so it also applies to previously published packages, rather than updating the link here.

It is deploying now, so should be sorted in the next 10 mins or so

PR #1325, opened 2024-12-11
chore: fix typos

JP-Ellis:
Fix a few typos in the repo


mefellows (approved):
Nice thanks! the CHANGELOG is auto-generated from the commit messages, so it's possible if we ever re-generate it from scratch it will revert these (but no harm in fixing them for now).

Issue #1372, opened 2025-02-17
CI - Applied ruleset for master branch impacts release process


YOU54F:
Similar to https://github.com/pact-foundation/pact_broker/issues/779

package is created and pushed to npm. 

updated version and changelog is expected to be pushed back, and release created, however new branch/ci restrictions impacts this from executing successfully

```
npm notice Publishing to https://registry.npmjs.org/ with tag latest and public access
+ @pact-foundation/pact@14.0.0
    done!
remote: error: GH013: Repository rule violations found for refs/heads/master.        
remote: Review all repository rules at https://github.com/pact-foundation/pact-js/rules?ref=refs%2Fheads%2Fmaster        
remote: 
remote: - Required status check "Test completion check" is expected.        
remote: 
To https://github.com/pact-foundation/pact-js
 * [new tag]           v14.0.0 -> v14.0.0
 ! [remote rejected]   master -> master (push declined due to repository rule violations)
error: failed to push some refs to 'https://github.com/pact-foundation/pact-js'
```

This has stopped

- Updated package.json / lockfile from being committed back
- Updated changelog from being committed back
- A release from being created, with the appropriate changelog

furthermore, renovate auto-merged a PR after the release, meaning it's awkward to retrospectively update the changelog and version in a sequential manner.

Ideally the release workflow should work as before the new restrictions were applied, or the release workflow should be able to circumvent restrictions.

cc: @JP-Ellis

JP-Ellis:
Do you know which account is being used to push to `master`? The rules can be adjusted to exempt that user specifically.

YOU54F:
Whichever user triggers the workflow_dispatch action.

This has caused us to release 15.0.0 instead of 14.0.1 :( 

comment - https://github.com/pact-foundation/pact-js/pull/1394#issuecomment-2748972705

Because we had to retrospectively add the version bump commit, that should have committed from GH actions, this has caused the changelog history to go awry, and meant the last release that has just gone out, went out as a major, rather than a 
patch.

changelog history between 14.0.0 and 15.0.0

https://github.com/pact-foundation/pact-js/compare/v14.0.0...master

note the 4th commit https://github.com/pact-foundation/pact-js/pull/1373 

[...]

PR #1373, opened 2025-02-17
chore(release): 14.0.0

YOU54F:
correcting changelog & version which failed to update in the last release.

relates to #1372 

can't push straight to master (neither can CI on release), due to restriction of test completion check.

CI is impacted, as release workflow is kicked off manually, and does not trigger the test workflow, and therefore completion check. 

the release workflow itself, does perform tests prior to release, so maybe these need to be hooked up to the test completion check 🤷🏾‍♂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,31 @@
 
 All notable changes to this project will be documented in this file. See [commit-and-tag-version](https://github.com/absolute-version/commit-and-tag-version) for commit guidelines.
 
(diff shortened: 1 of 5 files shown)


YOU54F:
jest-pact needs peer dep updating

https://github.com/pact-foundation/jest-pact/pull/343

nestjs-pact needs peer dep updating

https://github.com/pact-foundation/nestjs-pact/pull/89

Issue #1076, opened 2023-03-18
`MatchersV3.date` and `MatchersV3.time` do not use example value on the provider side


agross:
### Software versions

- **OS**: macOS 13.2.1
- **Consumer Pact library**: Pact JS 11.0.2
- **Provider Pact library**: [PactNet 4.4.0](https://www.nuget.org/packages/PactNet/4.4.0)
- **Node Version**: `v16.19.0`

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures
- [x] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [ ] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour

We use `MatchersV3.date` and `MatchersV3.time`. The docs state:

> String value that must match the provided date/time format string. [...] If the example value is omitted, a value will be generated using a Date/Time generator and the current system date/time.

The expected behavior is that the provider will receive the passed example value of `10:11:12`:

```ts
const x = MatchersV3.time("HH:mm:ss", "10:11:12")
```

### Actual behaviour

The provider receives a generated date/time value, despite examples being given.

The Pact JSON file also contains `generators` for such date or time matchers.

Seems to be related to #620 which fixed it for `MatchersV3.datetime`.

While researching this I came across [this document](https://docs.rs/pact_models/latest/pact_models/generators/datetime_expressions/index.html) which seems to indicate that date/time values on the provider side could be dynamically generated. How can such expressions be included when generated values are used (i.e. no concrete examples are given, but a recipe describing the values to be generated). It would be very useful to use because "the current date/time" is not enough for most of our use-cases.

### Steps to reproduce

```ts
const d = Matchers.date("yyyy-MM-dd", "2023-02-01");
expect(d["pact:generator:type"]).toBeUndefined();

const t = Matchers.time("HH:mm:ss", "10:11:12");
expect(t["pact:generator:type"]).toBeUndefined();
```

TimothyJones:
> The expected behavior is that the provider will receive the passed example value of 10:11:12:

No, it isn't. 

[...]

PR #1405, opened 2025-03-14
fix: mock backend to return the example value provided in date and time v3 matchers

pulak777:

### PR Template

Currently while using MatchersV3 `date` and `time`, mock backend returns current date instead of the provided example date. 

This works correctly for `datetime`. So this pr applies similar fix to [what was done in `datetime`](https://github.com/pact-foundation/pact-js/blob/master/src/v3/matchers.ts#L336) (removing `pact:generator:type` from output object when example value is present)

Added some tests in pact-reference for this scenario as well - https://github.com/pact-foundation/pact-reference/pull/489


pulak777:
Looks like there's been a discussion on this before - https://github.com/pact-foundation/pact-js/issues/1076

mefellows (approved):

mefellows:
Thanks!

pulak777:
Thank you @mefellows. From [this comment](https://github.com/pact-foundation/pact-js/issues/1076#issuecomment-1683503000) and the [commit ](https://github.com/pact-foundation/pact-js/commit/d04ab2020df9f97082160c224d9ace9b79d96cd4) where the example was made mandatory, it looks like we can make the example input optional. As rust ffi generates example value from the given regex (which probably was not the case then)

PR #1505, opened 2025-06-20
docs: update npm install command

zachlysobey:
### Description

This PR just updates the `npm i` command in the `README`.
- switches to longhand `install`
- removes `-S` (short for `--save`) which has been the default for a long time and is no longer necessary
- adds `--save-dev` (because the vast majority of the time, pact is used as dev-time, not as a production dependency)
- removes `@latest` which is redundant and unnecessary

```diff
-npm i -S @pact-foundation/pact@latest
+npm install --save-dev @pact-foundation/pact
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Learn the key Pact JS features in 60 minutes: https://github.com/pact-foundation
 ## Installation
 
 ```shell
-npm i -S @pact-foundation/pact@latest
+npm i @pact-foundation/pact@latest
 
 # 🚀 now write some tests!
 ```
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Learn the key Pact JS features in 60 minutes: https://github.com/pact-foundation
 ## Installation
 
 ```shell
-npm i @pact-foundation/pact@latest
+npm install --save-dev @pact-foundation/pact
 
 # 🚀 now write some tests!
 ```


YOU54F:
Would `--save-dev` or `-D` not be better. 

I would expect pact-js to be in the devDependencies for the majority of projects, unless you are building tooling on top of pact-js.

Where did you find that said `--save` was deprecated?

I see it listed as an option, just defaults to true now.

https://docs.npmjs.com/cli/v11/commands/npm-install#save

I agree being more verbose is better in a starting guide, as it helps newbies who may not understand what the short hand flags are

zachlysobey:
@YOU54F whelp, I was wrong about `--save` deprecation. I read that somewhere, but shoulda gone to the definitive source.
Also, I agree with your suggestion about `--save-dev`. I'll make the changes as suggested

README.md
```diff
@@ -91,7 +91,7 @@ Learn the key Pact JS features in 60 minutes: https://github.com/pact-foundation
 ## Installation
 
 ```shell
-npm i -S @pact-foundation/pact@latest
+npm i @pact-foundation/pact@latest
```
zachlysobey:
```suggestion
npm install --save-dev @pact-foundation/pact
```

YOU54F:
all good, cheers for the change dude

PR #1513, opened 2025-06-25
docs: fix provider stateHandlers example

zachlysobey:
## What does this PR do?

This PR makes a couple small changes to the [Provider Verification docs](https://docs.pact.io/implementation_guides/javascript/docs/provider#api-with-provider-states) – specifically the *API with Provider States* `stateHandlers` example code

This fixes two issues:
1) there was a missing comma, so the example was technically broken
2) `[null]:` as a property key doesn't work in TypeScript. This updates it to work in JS *and* TS

```diff
 const opts = {
   ...
   stateHandlers: {
-    [null]: () => {
+    "null": () => {
       // This is the "default" state handler, when no state is given
-    }
+    },
     "Has no animals": () => {
       animalRepository.clear()
       return Promise.resolve(`Animals removed from the db`)
```

## More on `[null]`

When this code is used with TS, it raises the following error:
> `A computed property name must be of type 'string', 'number', 'symbol', or 'any'`

TS is more strict than JS here, which allows it, but I'm pretty sure `[null]` is just coerced into `"null"` at runtime with JS, so I put that in explicitly.

### Slack thread

I raised this in [the Pact Slack](https://pact-foundation.slack.com/archives/C9VBGLUM9/p1750801201014229) yesterday, but figured I'd go ahead and just cut a PR to get the ball rolling

#### misleading/wrong attribution of example

Just above this code example, it says "Here is an example from our [e2e suite](https://github.com/pact-foundation/pact-js/blob/master/examples/e2e/test/provider.spec.js):". But the `[null]` case is not actually part of that example, and looks like it was added later (although I couldn't easily follow the Git history with `blame`).

I think it'd be best to resolve this, to avoid any confusion. Perhaps just removing that line?

#### thoughts on a different approach to registering state handlers

I tend to think a `getStateHandler` function would  have been a better pattern than `stateHandlers` for this, b/c it'd allow stuff like switch statements, which handle default cases more elegantly

```
getStateHandler: (stateName: string) {
  switch (stateName) {
    case "Has no animals":
      return () => { ... },
    default:
      return () => { ... },
  }
}
```

Maybe there are compelling reasons to have it as it is with the static `stateHandlers` object though 🤷

### DIFF
--- a/docs/provider.md
+++ b/docs/provider.md
@@ -120,9 +120,9 @@ If you have defined any `state`s in your consumer tests, the `Verifier` can put 
 const opts = {
   ...
   stateHandlers: {
-    [null]: () => {
+    "null": () => {
       // This is the "default" state handler, when no state is given
-    }
+    },
     "Has no animals": () => {
       animalRepository.clear()
       return Promise.resolve(`Animals removed from the db`)


zachlysobey:
CI here is failing, but presumably not due to anything I did
```
> e2e@1.0.0 test:publish
> pact-broker publish ./pacts --consumer-app-version="$(npx absolute-version)" --auto-detect-version-properties --broker-base-url=https://testdemo.pactflow.io/

PactBroker::Client::Hal::ErrorResponseReturned - Error making request to https://testdemo.pactflow.io/ status=401 Unauthorized. Please see https://docs.pactflow.io/docs/login-help for more information.
Error: Process completed with exit code 1.
```

mefellows (approved):

docs/provider.md
```diff
@@ -120,9 +120,9 @@ If you have defined any `state`s in your consumer tests, the `Verifier` can put
 const opts = {
   ...
   stateHandlers: {
-    [null]: () => {
+    "null": () => {
```
mefellows, line 123: hmm this actually just matches the string `"null"`, and not a `null` handler. Perhaps we're better updating the types to support this more directly?

zachlysobey: In JS, all object property keys (except for `Symbols`) are `string`s.  When you use `[null]` like this, it coerces it to a string under the hood.

A few examples that demonstrate this

```js
const obj = {};
obj[null] = 42;
console.log(obj); // { "null": 42 }
console.log(obj["null"]); // 42
```

```js
Object.keys({ [null]: 1 }) // ['null']
Object.keys({ "null": 1 }) // ['null']
```

```js
const obj1 = { [null]: 1, "null": 2 }
const obj2 = { "null": 2, [null]: 1 }
JSON.stringify(obj1) // '{"null":2}'
JSON.stringify(obj2) // '{"null":1}'
```

zachlysobey:
> Yeah, I think that's because you're an external contributor. We need to separate that part of the pipeline out I think (or find an alternative solution).

I did just succesfully make another (doc-only) PR via #1505, and it did not run into this issue

docs/provider.md
zachlysobey: All of those examples are how it works with the node REPL, or the browser console.
If you run those examples with `ts-node` (the TypeScript REPL) they all fail with runtime errors.
Compiling with `tsc` also fails for the same reason.
> `A computed property name must be of type 'string', 'number', 'symbol', or 'any'`

This isn't a matter of _updating the types_ or something. It's how TS was designed. They identify this pattern as confusing and problematic, and explicitly disallow it.

_[TS1170](https://lior-amsalem.hashnode.dev/ts1170-a-computed-property-name-in-a-type-literal-must-refer-to-an-expression-whose-type-is-a-literal-type-or-a-unique-symbol-type) has a bit more context_

mefellows:
Thanks for the explanation! 

I like your `getStateHandler` suggestion, I think it would more neatly allow this kind of pattern:

```
getStateHandler: (stateName: string) {
  if (stateName.match(someRegex) {
     // do something
  } else if (...)
    // do some other thing
  } else {
    // do fallback
  }
}
```
Which could make provider state handling even more expressive. 

Alternatively, an explicit "fallback handler" in the type could be nice as well, allowing for the above "catch anything" scenario. Happy to explore some ideas here through a feature request!

PR #1582, opened 2025-10-15
fix: Pact-JS Verifier Proxy Cannot Handle Root Path

### LINKED ISSUE #1581: Pact-JS Verifier Proxy Cannot Handle Root Path
## Summary

The Pact-JS verifier's internal proxy server cannot forward requests to the root path `/` due to an incorrect route pattern in the proxy configuration. This prevents provider verification from working when the API endpoint is at the root path.

## Environment

- **Pact-JS Version**: 16.0.0
- **Node.js Version**: Tested on multiple versions
- **Express Version**: 5.1.0 (used by Pact-JS)
- **Operating System**: macOS (but likely affects all platforms)

## Problem Description

When using the Pact-JS `Verifier` with a provider that serves API endpoints at the root path `/`, the verifier fails with a 404 error. The issue is in the proxy server's route configuration.

### Expected Behavior

The verifier should be able to forward requests to any path, including the root path `/`.

### Actual Behavior

The verifier returns a 404 "Cannot POST /" error when trying to verify providers that serve endpoints at `/`.

## Root Cause Analysis

The issue is in the Pact-JS source code at `src/dsl/verifier/proxy/proxy.ts` line 86:

```typescript
// src/dsl/verifier/proxy/proxy.ts (line 86)
app.all('/*splat', (req, res) => {
  logger.debug(`Proxying ${req.method}: ${req.path}`);

  proxy.web(req, res, toServerOptions(config, req));
});
```

The route pattern `/*splat` does **not** match the root path `/` in Express. This is because:

1. `/*` matches `/` ✓
2. `/*splat` does NOT match `/` ✗ (requires at least one character in the parameter)
3. `/{splat}` DOES match `/` ✓

### Evidence

Testing with Express shows the difference:

```javascript
// This works - matches /
app.all('/*', (req, res) => { res.send('matched'); });

// This works - matches /
app.all('/{splat}', (req, res) => { res.send('matched'); });

// This does NOT work - doesn't match /
app.all('/*splat', (req, res) => { res.send('matched'); });
```

### Inconsistency in Codebase

Interestingly, the same repository has the correct pattern in `src/dsl/verifier/proxy/messages.ts` (line 162):

```typescript
// src/dsl/verifier/proxy/messages.ts (line 162) - CORRECT
app.all('/{splat}', createProxyMessageHandler(config));
```

This shows that the correct pattern is known and used elsewhere in the codebase.

### 1. Consumer Test (creates the pact)

```typescript
// consumer/test.spec.ts
import { PactV4, LogLevel, SpecificationVersion, V3MockServer } from '@pact-foundation/pact';
import * as path from 'path';

describe("HTTP Provider Test", () => {
    it("should handle root path", async () => {
        const pact = new PactV4({
            consumer: 'test-consumer',
            provider: 'test-provider',
            logLevel: 'info' as LogLevel,
            spec: SpecificationVersion.SPECIFICATION_VERSION_V4,
        });

        await pact
            .addInteraction()
            .given('a request')
            .uponReceiving('a request to root path')
            .withRequest('POST', '/', (builder) => {
                builder.headers({'Content-Type': 'application/json'});
                builder.jsonBody({message: 'test'});
            })
            .willRespondWith(200, (builder) => {
                builder.headers({ 'Content-Type': 'application/json' });
                builder.jsonBody({status: 'success'});
            })
            .executeTest(async (mockserver: V3MockServer) => {
                const response = await fetch(`http://localhost:${mockserver.port}/`, {
                    method: 'POST',
                    headers: { 'Content-Type': 'application/json' },
                    body: JSON.stringify({message: 'test'})
                });
                const data = await response.json();
            });
    });
});
```

### 2. Provider Test (fails with current bug)

```typescript
// provider/test.spec.ts
import { Verifier } from '@pact-foundation/pact';
import express from 'express';

const app = express();
app.use(express.json());

// Provider endpoint at root path
app.post('/', (req, res) => {
    res.json({status: 'success'});
});

describe('Provider Verification', () => {
    let server: any;
    const PORT = 8080;

    beforeAll(async () => {
        server = app.listen(PORT);
        await new Promise(resolve => setTimeout(resolve, 1000));
    });

    afterAll(async () => {
        if (server) server.close();
    });

    it('should verify provider', async () => {
        const opts = {
            provider: 'test-provider',
            providerBaseUrl: `http://localhost:${PORT}`,
            pactUrls: [path.resolve(process.cwd(), 'pacts/test-consumer-test-provider.json')],
            stateHandlers: {
                'a request': () => Promise.resolve()
            }
        };

        await new Verifier(opts).verifyProvider();
    });
});
```

### 3. Expected vs Actual Results

**Expected**: Test passes, provider verification succeeds.

**Actual**: Test fails with:
```
Verification failed
Failures:
1) has status code 200
   expected 200 but was 404
```

## Workaround

The current workaround is to avoid using the root path `/` and instead use a subpath like `/api`:

```typescript
// In consumer test
.withRequest('POST', '/api', ...)

// In provider test  
app.post('/api', (req, res) => { ... });
```

## Proposed Fix

Change line 86 in `src/dsl/verifier/proxy/proxy.ts` from:

```typescript
app.all('/*splat', (req, res) => {
```

to:

```typescript
app.all('/{splat}', (req, res) => {
```

## Additional Context

1. **Inconsistency**: The `messages.ts` file in the same directory correctly uses `'/{splat}'` (line 162), which works properly.

2. **Express Version**: This issue affects Express 5.x. The route pattern behavior may have changed between Express versions.

3. **Impact**: This bug affects any provider that serves APIs at the root path, which is a common pattern for simple APIs.

## Files Affected

- `src/dsl/verifier/proxy/proxy.ts` (line 86) - **Primary file with the bug**
- `src/dsl/verifier/proxy/messages.ts` (line 162) - **Reference file showing correct pattern**
- Potentially other files that use the same incorrect pattern

## Priority

**High** - This is a fundamental issue that prevents provider verification from working with a common API pattern (root path endpoints).

## Additional Notes

The debug logs show that the proxy receives the request but doesn't forward it:

```
[DEBUG] incoming request: {"method":"POST","path":"/"}
[DEBUG] outgoing response: {"status":404,"body":"Cannot POST /"}
```

This confirms that the route pattern `/*splat` is not matching the root path `/`.

YOU54F:
update express wildcard path

fixes #1581

### DIFF
--- a/src/dsl/verifier/proxy/proxy.ts
+++ b/src/dsl/verifier/proxy/proxy.ts
@@ -89,7 +89,7 @@ export const createProxy = (
     proxy.web(req, res, toServerOptions(config, req));
   });
 
-  app.all('/'', (req, res) => {
+  app.all('/', (req, res) => {
     logger.debug(`Proxying ${req.method}: ${req.path}`);
 
     proxy.web(req, res, toServerOptions(config, req));


mefellows:
Hmm some tests failures - need to test locally to see if they're related to the change or not.

YOU54F:
I am pretty sure these are the same failures I saw when upgrade express to 5.x in the upgrade all the things PR, hence why there was a different splat pattern in the two places.

maybe we need two routes, one for the root path and one for any wildcard route, to cover both cases?

mefellows:
Thanks! Looking at the docs, wondering if this would work: https://expressjs.com/en/guide/migrating-5.html#path-syntax

Screenshot 2025-10-16 at 12 48 39 pm

But if this works, that's cool too!

Issue #1140, opened 2023-11-28
Not possible to use multipart form data because of boundary


jawadsef:
### Software versions

_Please provide at least OS and version of pact-js_

- **OS**: `macOS Ventura Version 13.0.1 (22A400)`
- **Consumer Pact library**: _e.g. Pact JS v2.6.0_
```
"@pact-foundation/pact": "^12.1.0"
"@pact-foundation/pact-core": "^14.0.5"
```
- **Provider Pact library**: not relevant, since this is just a consumer test
- **Node Version**: `v18.14.1`

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code and have [read](https://github.com/pact-foundation/pact-js/blob/master/docs/troubleshooting.md#test-fails-when-it-should-pass) the section on intermittent test failures
- [x] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [x] For bonus points and virtual high fives, I have created a reproducible git repository (see below) to illustrate the problem

### Expected behavior

Test passes.

### Actual behavior

Test fails.

Getting the following error:
```
Mock server failed with the following mismatches:
        0) The following request was incorrect: 
                POST /test            
                         1.0    $: Failed to parse the expected body as a MIME multipart body: 'no boundary in content-type'
                         1.1    $: Failed to parse the actual body as a MIME multipart body: 'no boundary in content-type'
```

I tried generating the boundary but didn't work either.

### Steps to reproduce

Clone following repo: https://github.com/jawadsef/pact-js-repro-template (`git clone git@github.com:jawadsef/pact-js-repro-template.git`)

Run consumer test: https://github.com/jawadsef/pact-js-repro-template/blob/main/consumer.spec.ts (`npm run test:consumer  `)

### Relevant log files

```
[23:34:01.279] INFO (41652): 0.4.0: pact native library successfully found, and the correct version
2023-11-28T22:34:01.331419Z DEBUG ThreadId(01) pact_plugin_driver::catalogue_manager: Updated catalogue entries:
core/transport/http
core/transport/https
2023-11-28T22:34:01.333204Z DEBUG ThreadId(01) pact_plugin_driver::catalogue_manager: Updated catalogue entries:
core/content-generator/binary
core/content-generator/json
core/content-matcher/json
core/content-matcher/multipart-form-data
core/content-matcher/text
core/content-matcher/xml
[...]

PR #1627, opened 2025-12-11
chore: update types and add examples for multipart usage

kevinrvaz:
Add examples for multipart form usage, and binding changes for https://github.com/pact-foundation/pact-js-core/pull/782
also addresses - https://github.com/pact-foundation/pact-js/issues/1140


kevinrvaz:
build issue will get fixed once https://github.com/pact-foundation/pact-js-core/pull/782 package update is added here

examples/v3/multipart/consumer.spec.ts
JP-Ellis, line 1: Since this is an example, and therefore is meant to be a reference that other people can use to learn how to use Pact-JS, this should have a fairly decent blurb, both in a `README.md` for some of the higher-level explanations, and within comments here (with a certain degree of overlap, because repetition helps).

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+  });
+
+  it('creates a pact to verify with a text field in form', async () => {
+    // Create a multipart form data object with a 'name' field
+    const formData = new FormData();
+    formData.append('name', 'John Doe');
+
+    // Extract the headers that include the Content-Type with the boundary parameter
+    // This boundary is crucial for multipart form data parsing
+    const headers = formData.getHeaders();
```
JP-Ellis: In many cases, the end-user would probably want to add additional headers. Additionally, it is important to set up matching rules that ensure the arbitrary boundaries can be matched, as opposed to the one specific generated in this case.

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import {
+  SpecificationVersion,
+  PactV3,
+  MatchersV3,
+} from '@pact-foundation/pact';
+import FormData from 'form-data';
```
JP-Ellis, line 26: Is this a node-special? Or a library?

If the former, this should probably imported `from 'node:form-data'`.

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+    // Create a multipart form data object with a 'name' field
+    const formData = new FormData();
+    formData.append('name', 'John Doe');
+
+    // Extract the headers that include the Content-Type with the boundary parameter
+    // This boundary is crucial for multipart form data parsing
+    const headers = formData.getHeaders();
+
+    // Convert the form data to a string representation for the request body
+    const form = formData.getBuffer().toString();
```
JP-Ellis, line 61: Is it not possible to pass this in directly to the `body` parameter? Just double checking that `formData` doesn't have a seamless 'stringify' compatibility for example.

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    await pact
+      .addInteraction({
+        uponReceiving: 'a request for a foo',
```
JP-Ellis: It would be nice to avoid 'foo' and 'bar' in this example, and instead having a semi-plausible interaction.

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+  MatchersV3,
+} from '@pact-foundation/pact';
+import FormData from 'form-data';
+import axios from 'axios';
+
+chai.use(chaiAsPromised);
+
+const { expect } = chai;
+
+describe('Pact Consumer Test', () => {
```
mefellows: One example that we're missing is the use of _multiple_ multipart bodies (as per [this comment](https://github.com/pact-foundation/pact-js/issues/1140#issuecomment-2989602877)). Could you please add another scenario showing how to add multiple? Or perhaps just add multiple to the existing ones to avoid the need  for too many test cases)

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+          method: 'POST',
+          path: '/test',
+          body: form, // The multipart form data as a string
+          headers, // Headers including Content-Type with boundary
+        },
+        willRespondWith: {
+          status: 201,
+          body: {
+            // Use a matcher to flexibly match any string value for 'foo'
+            foo: MatchersV3.like('bar'),
```
JP-Ellis: It would be nice to have a more plausible example. For example, having the response be relevant to the input. In the Pact Python example, I'm using an example of uploading a JPEG file along with metadata for the file; and then the response returning the size of the JPEG.

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+        uponReceiving: 'a request for a foo',
+        withRequest: {
+          method: 'POST',
+          path: '/test',
+          body: form, // The multipart form data as a string
+          headers, // Headers including Content-Type with boundary
+        },
+        willRespondWith: {
+          status: 201,
+          body: {
```
JP-Ellis, line 74: Ultimately, it would be good to have matching rules on the body itself, but I don't think that is possible at this stage.

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+        // Verify the response indicates success
+        expect(response.data.success).to.equal(true);
+        expect(response.data.message).to.equal('File uploaded successfully');
+
+        // Clean up the temporary file
+        fs.unlinkSync(tmpFile);
+      });
+  });
+
+  it('creates a pact using withRequestMultipartFileUpload helper method', async () => {
```
JP-Ellis, line 198: Why do we have both ways of doing this? I might be missing the context, but this example should probably explain this in some detail.

examples/v3/multipart/consumer.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+        );
+
+        // Verify the response
+        expect(response.data.success).to.equal(true);
+        expect(response.data.filename).to.equal('test-upload-file.txt');
+        expect(response.data.message).to.equal('File received successfully');
+      });
+  });
+
+  it('creates a pact using withRequestMultipartFileUpload with custom boundary', async () => {
```
JP-Ellis, line 247: And this example should explain in what circumstances we would want a custom boundary, pros and cons, etc.

src/v3/pact.ts
JP-Ellis, line 199: Do we need to update docstrings somewhere? (I don't know where they are)

examples/v3/multipart/consumer.spec.ts
kevinrvaz: this is node specific, ill update this.

src/v3/pact.ts
kevinrvaz: i didn't see any docstrings for any of the class methods either, but i can perhaps add them for the functions i am touching.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: i was showing multiple ways of how one can pass multipart form data, either directly with normal fields or file fields, or built in methods like withRequestMultipartFileUpload with and without boundaries.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: ill add a jpeg eg as well

examples/v3/multipart/consumer.spec.ts
kevinrvaz: i had adapted it originally from - https://github.com/kevinrvaz/pact-js-repro-template/blob/main/consumer.spec.ts, that's also where the foo bar came from, which is based on the original issue that the github issue listed. The form if passed directly in a withRequest method has issues, but you can definitely pass it directly in execute test.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: ill add a read me as well as more details for each test, i was thinking the comments i added was enough, but ill add more.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: ill add a read me as well as more details for each test, i was thinking the comments i added was enough, but ill add more.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: i can add more headers, to show you can pass additional headers.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: after checking some more this lib is also can be used in the browser, so leaving it as is.

src/v3/pact.ts
kevinrvaz: added

examples/v3/multipart/consumer.spec.ts
JP-Ellis: That's fine, and actually good to show different ways this can be done. It just wasn't clear from looking at the file alone _why_ there are multiple ways of doing this.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: i've added a more detailed readme on how to run the tests and what each of it does, as well as more comments

examples/v3/multipart/consumer.spec.ts
kevinrvaz: ive added more details in the readme

examples/v3/multipart/consumer.spec.ts
kevinrvaz: added in readme

examples/v3/multipart/consumer.spec.ts
kevinrvaz: added.

examples/v3/multipart/consumer.spec.ts
kevinrvaz: added

examples/v3/multipart/consumer.spec.ts
kevinrvaz: added

mefellows (approved):

PR #1628, opened 2025-12-11
fix: replace deprecated macos 13 build with macos-15-intel

kevinrvaz:
Remove deprecated macos build - https://github.com/actions/runner-images/issues/13046 which is breaking for https://github.com/pact-foundation/pact-js/pull/1627

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,6 +98,7 @@ jobs:
       matrix:
         node-version: [20, 22, 24]
         os:
+          - macos-15-intel
           - macos-latest
           - ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -98,7 +98,6 @@ jobs:
       matrix:
         node-version: [20, 22, 24]
         os:
-          - macos-13
```
JP-Ellis, line 101: We shouldn't _remove_ the intel runners, we should instead upgrade this to `macos-15-intel`.

kevinrvaz: changed

PR #1640, opened 2026-01-07
feat: add support for with_matching_rules and add example usage

kevinrvaz:
Added support for with_matching_rules binding changes of https://github.com/pact-foundation/pact-js-core/pull/791
also addresses - https://github.com/pact-foundation/pact-js/issues/1140

### DIFF
--- a/src/v3/pact.ts
+++ b/src/v3/pact.ts
@@ -15,7 +15,6 @@ import {
   V3Interaction,
   V3MockServer,
   V3ProviderState,
-  V3RegexMatcher,
   V3Request,
   V3Response,
 } from './types';


examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+const { expect } = chai;
+
+describe('Pact Multipart with Matching Rules', () => {
+  /**
+   * Initialize a new Pact instance with consumer and provider names.
+   */
+  const pact = new PactV3({
+    consumer: 'multipart-consumer',
+    provider: 'multipart-provider',
+    spec: SpecificationVersion.SPECIFICATION_VERSION_V3,
```
JP-Ellis: Is this needed when using `PactV3`?

examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+    expectedBody += `Content-Type: image/jpeg\r\n`;
+    expectedBody += `\r\n`;
+
+    // Convert string to buffer and append binary JPEG data
+    let bodyBuffer = Buffer.from(expectedBody);
+    bodyBuffer = Buffer.concat([bodyBuffer, JPEG_BYTES]);
+    bodyBuffer = Buffer.concat([
+      bodyBuffer,
+      Buffer.from(`\r\n--${boundary}--\r\n`),
+    ]);
```
JP-Ellis: Nit: We can probably just have one `Buffer.concat` call with three parts:

```ts
let bodyBuffer = Buffer.concat([
  Buffer.from(expectedBody),
  JPEG_BYTES,
  Buffer.from(`\r\n--${boundary}--\r\n`),
])
```

examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+        /**
+         * Execute the test with different data that still matches the rules.
+         *
+         * Note that we're sending:
+         * - Different name: "different" instead of "test" (but still matches [a-zA-Z]+)
+         * - Different size: 200 instead of 100 (but still an integer)
+         * - Same JPEG data (matches image/jpeg content type)
+         *
+         * The matching rules ensure this different data is still valid.
+         * We use the same fixed boundary to ensure the request matches.
```
JP-Ellis: Is that still necessary? A lot of libraries don't allow for the boundary to be specified (hence the need for matching rules).

examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+     *
+     * Matching rules for multipart bodies use the format:
+     * - $.image - matches the entire image part (for content type)
+     * - $.metadata - matches the entire metadata part (for type)
+     * - $.metadata.name - matches a field within the metadata JSON
+     * - $.metadata.size - matches another field within the metadata JSON
+     *
+     * These rules allow flexibility in the actual data sent while
+     * ensuring the contract structure is maintained.
+     */
```
JP-Ellis: Typically, the best practice is to use matching rules within the JSON payload (using the dedicated classes). This is not supported within the multipart body, hence the need for matching rules being specified separately. This should be mentioned more clearly.

examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+    const requestMatchingRules = {
+      body: {
+        // Match content type of the image part to allow any valid JPEG
+        '$.image': {
+          matchers: [
+            {
+              match: 'contentType',
+              value: 'image/jpeg',
+            },
+          ],
```
JP-Ellis: Is it possible to use the TS classes for matches here? Otherwise, people won't know the syntax.

examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+
+    await pact
+      .given('file upload service is available')
+      .uponReceiving('a multipart upload with JSON metadata and image')
+      .withRequestMatchingRules(
+        {
+          method: 'POST',
+          path: '/upload',
+          headers: {
+            'Content-Type': `multipart/form-data; boundary=${boundary}`,
```
JP-Ellis: There should be a matcher for the boundary here (which resolves an issue below).

examples/v3/multipart/README-matching-rules.md
JP-Ellis, line 1: Why is this file not just called `README.md`? Or incorporated into an existing `README` (assuming that's the reason why it was named differently).

examples/v3/multipart/README-matching-rules.md
```diff
@@ -0,0 +1,299 @@
...
+
+```bash
+# Run only the multipart matching rules tests
+npm run test:matching-rules
+
+# Run the regular multipart tests
+npm run test:consumer
+```
+
+## Test Cases Explained
```
JP-Ellis: I would abbreviate a lot of the explanation here, and instead direct the user to focus on the code itself. A lot of what you say here is duplicated from the comments/docstrings, and you also duplicate a lot of code into the README.

examples/v3/with-matching-rules/consumer.spec.ts
```diff
@@ -0,0 +1,501 @@
...
+const { expect } = chai;
+
+describe('Pact Consumer Test Using Matching Rules', () => {
+  /**
+   * Initialize a new Pact instance with consumer and provider names.
+   */
+  const pact = new PactV3({
+    consumer: 'matchingrulesconsumer',
+    provider: 'matchingrulesprovider',
+    spec: SpecificationVersion.SPECIFICATION_VERSION_V3,
```
JP-Ellis: As above, is this needed?

examples/v3/with-matching-rules/README.md
```diff
@@ -0,0 +1,279 @@
+# Pact JS V3 Matching Rules Example
+
+This example demonstrates how to use Pact JS V3 with matching rules, which allow you to define flexible matching criteria for request and response attributes beyond exact equality.
```
JP-Ellis: I might stand corrected here, but Pact JS already has supported for matching rules through the so-called ['integration JSON'](https://docs.pact.io/implementation_guides/rust/pact_ffi/integrationjson). This allows for rules to be embedded directly within JSON payloads, and I'm pretty sure Pact-JS has a number of classes to support this already (as opposed to having people specify the JSON explicitly).

The main purpose of the matching rules FFI is to allow for matching rules to be specified in circumstances where the integration JSON is not supported (e.g., multipart form data).

examples/v4/with-matching-rules/README.md
```diff
@@ -0,0 +1,537 @@
...
+      .headers({ 'Content-Type': 'application/json' })
+      .jsonBody({
+        email: 'jessica.martinez@acmecorp.com',
+        phone: '+14155552671',
+      })
+      .matchingRules(JSON.stringify(requestMatchingRules));
+  })
+  .willRespondWith(201, (builder) => {
+    builder.jsonBody({
+      id: Matchers.uuid('a3f2c8b1-9d4e-4c7a-b2e5-f8a9c1d3e5f7'),
```
JP-Ellis: I see you have the Matcher here... does that mean the v3 doesn't support these?

examples/v4/with-matching-rules/README.md
```diff
@@ -0,0 +1,537 @@
...
+```typescript
+const requestMatchingRules = {
+  body: {
+    '$.email': {
+      matchers: [
+        {
+          match: 'regex',
+          regex: '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$',
+        },
+      ],
```
JP-Ellis: Can we add support for matchers here using the `Matchers.*` functions/methods?

src/v3/pact.ts
```diff
@@ -133,6 +133,52 @@ export class PactV3 {
     return this;
   }
 
+  /**
+   * Applies matching rules to the consumer request.
+   * Matching rules allow you to define flexible matching criteria for request attributes
+   * beyond exact equality (e.g., regex patterns, type matching, number ranges).
```
JP-Ellis, line 141: The main reason we have this is to allow matching rules to be specified in cases where the integration JSON cannot be used.

src/v3/pact.ts
```diff
@@ -133,6 +133,52 @@ export class PactV3 {
...
+   *
+   * @param req - The request configuration (method, path, headers, etc.)
+   * @param rules - The matching rules as a Map or JSON string. Rules should follow the Pact matching rules format.
+   * @returns The PactV3 instance for method chaining
+   */
+  public withRequestMatchingRules(
+    req: V3Request,
+    rules: Map<string, unknown> | string
+  ): PactV3 {
```
JP-Ellis: How does this combine with other `withRequest` methods? For example, how would I use `withRequestBinaryFile` and also specify a matching rule?

In Python, the matching rule method is stand-alone and does just that. Would that be possible here? Or is that a v4-only special?

examples/v3/multipart/consumer-matching-rules.spec.ts
kevinrvaz: removed

examples/v3/multipart/consumer-matching-rules.spec.ts
kevinrvaz: updated.

examples/v3/with-matching-rules/consumer.spec.ts
kevinrvaz: removed.

examples/v4/with-matching-rules/README.md
kevinrvaz: v3 does have a uuid matcher

examples/v3/multipart/README-matching-rules.md
kevinrvaz: there is another readme which just focuses on the withmultipart form functions and not the with matching rules one, i can consolidate both.

src/v3/pact.ts
kevinrvaz: the methods can be chained together in both v3 and v4, as we return the request/response builder instance.

kevinrvaz: as discussed will replace this method by enhancing existing methods instead of having a separate method.

examples/v3/with-matching-rules/README.md
mefellows: that's correct

examples/v3/multipart/README-matching-rules.md
mefellows: Yeah. There's lots of really good stuff in here that might be best elevated to our `docs` section (which would then appear on docs.pact.io with minimal or no work). Perhaps extracting the multipart strategies and custom matching rules into this guide: https://github.com/pact-foundation/pact-js/blob/master/docs/matching.md as a starting point?

Then the specifics of how and what these examples do could be merged into the README in this subfolder.

examples/v3/multipart/consumer-matching-rules.spec.ts
mefellows: Yes, if we can create types that describe what comprises a matching rule, that would be helpful.

examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+          },
+          body: bodyBuffer.toString('binary'),
+        },
+        JSON.stringify(requestMatchingRules)
```
mefellows: Having to serialise an untyped JSON blob here to put into the DSL isn't the nicest API. I believe separating the rules application into its own method (suggested elsewhere here, similar to Pact Python) would make this a lot cleaner.

examples/v3/multipart/consumer-matching-rules.spec.ts
mefellows: Yep, this is one of the problems identified in the linked issue(s). If we can avoid hard-coding the boundary, that would be better (in most cases, this would be the better default to show anyway to avoid people running into the issue of dynamic boundaries, and then figuring out how to make it static)

src/v3/matchers.ts
```diff
@@ -294,6 +315,22 @@ export function string(str = 'some string'): Matcher<string> {
   };
 }
 
+/**
+ * Matches the content type of a multipart field.
+ * Used for matching binary content or specific content types in multipart requests.
+ * @param contentTypeValue The content type to match (e.g., 'image/jpeg', 'text/plain')
+ */
+export const contentType = (contentTypeValue: string): {
```
mefellows: This returns a list of matchers, unlike the others. I think for this module, it should return a consistent interface (i.e. returns a `Matcher`). 

I realise this was introduced for use with the new "with matching rules" capability, but we can consider a separate utility function for this (or, ideally, the new `withMatchingRules` API accepts and formats the correct shape).

src/v3/matchers.ts
```diff
@@ -216,48 +216,69 @@ export const boolean = (b = true): Matcher<boolean> => ({
...
  */
-export const integer = (int?: number): Matcher<number> => {
-  if (Number.isInteger(int)) {
-    return {
-      'pact:matcher:type': 'integer',
-      value: int,
-    };
-  }
-  if (int) {
+export const integer = (
```
mefellows: What was the reason for updating this and the `decimal` matcher?

kevinrvaz: inorder to use min and max
Screenshot 2026-01-15 at 9 59 54 AM

src/v3/matchers.ts
kevinrvaz: ill update it to just return a matcher

src/v3/matchers.ts
kevinrvaz: ill remove this, it doesnt seem to be supported with integer and decimal. Since the json wasn't erroring out i had assumed this is working

examples/v3/multipart/consumer-matching-rules.spec.ts
kevinrvaz: eg doesnt set any boundary anymore

examples/v3/multipart/consumer-matching-rules.spec.ts
kevinrvaz: using types now

examples/v3/multipart/consumer-matching-rules.spec.ts
kevinrvaz: using a matcher for boundary now

examples/v3/multipart/consumer-matching-rules.spec.ts
kevinrvaz: added separate types as well as runtime validation of types.

examples/v4/with-matching-rules/consumer.spec.ts
```diff
@@ -0,0 +1,280 @@
+/**
```
mefellows, line 1: It's not entirely clear to me why we need 4 tests here - I can't quite see the benefit over just having one example for request matching and another for response. Suggest we keep it smaller (unless of course there is a good reason).

src/v3/index.ts
```diff
@@ -8,6 +8,20 @@ export * from './types';
  */
 export * as MatchersV3 from './matchers';
 
+/**
+ * Individual matcher function exports for direct use
+ */
+export {
```
mefellows: Are these strictly necessary? Or were they added for convenience?

src/v3/types.ts
```diff
@@ -50,6 +50,29 @@ export interface RulesMatcher<T> extends Matcher<T> {
...
+export type Part = 'path' | 'body' | 'header' | 'query' | 'status';
+
+export type JSONPath = string;
+
+/**
+ * A matching rule with optional JSONPath and array of matchers
+ */
+export type Rule = {
+  path?: JSONPath;
+  rule: Matcher<unknown>[];
```
mefellows: Suggest we refactor this to be `rules` (as it accepts multiple matchers, correct?).

src/v3/pact.ts
```diff
@@ -133,6 +136,59 @@ export class PactV3 {
     return this;
   }
 
+  public withRequestHeader(
```
mefellows: Let's add a doc string here to clarify when this might be used.

I'm a little concerned about leaking `index` to this layer of the API, which is quite low level. I'll share my wider thoughts in a comment.

examples/v3/multipart/consumer-matching-rules.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+      headers: {
+        'Content-Type': `multipart/form-data; boundary=${boundary}`,
+      },
+    };
+
+    try {
+      await pact
+        .given('file upload service is available')
+        .uponReceiving('a multipart upload with JSON metadata and image')
+        .withRequestHeader(
```
mefellows: Could this also be expressed using the `withRequestMatchingRules` function? If so, I suggest we remove this function altogether.

src/v3/index.ts
kevinrvaz: i had added them for convenience, but i can remove them

examples/v4/with-matching-rules/consumer.spec.ts
kevinrvaz: initially added it to show multiple product cases where you can use with matching rules, ill just keep one now.

kevinrvaz: removed

src/v3/types.ts
kevinrvaz: updated

examples/v3/multipart/consumer-matching-rules.spec.ts
kevinrvaz: updated

src/v3/pact.ts
kevinrvaz: this function has been removed.

src/v3/index.ts
kevinrvaz: updated

PR #1663, opened 2026-02-10
feat: add matching rules support to async and sync interactions

kevinrvaz:
Adding binding changes for https://github.com/pact-foundation/pact-js-core/pull/832 to enable using matching rules in sync and async messages for github issue - https://github.com/pact-foundation/pact-js/issues/1140

### DIFF
--- a/examples/v4/multipart/consumer-matching-rules.spec.ts
+++ b/examples/v4/multipart/consumer-matching-rules.spec.ts
@@ -170,6 +170,31 @@ describe('Pact V4 Multipart with Matching Rules', () => {
       ],
     };
 
+    const responseMatchingRules: Rules = {
+      body: [
+        {
+          path: '$.id',
+          rules: [Matchers.like('upload-1')],
+        },
+        {
+          path: '$.message',
+          rules: [Matchers.like('Upload successful')],
+        },
+        {
+          path: '$.metadata.name',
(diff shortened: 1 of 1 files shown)


src/v4/message/index.ts
```diff
@@ -149,6 +151,13 @@ export class SynchronousMessageWithRequestBuilder
 
     return this;
   }
+
+  matchingRules(rules: Rules): V4SynchronousMessageWithRequestBuilder {
```
mefellows: Is there a reason you didn't  go with `withMatchingRules` to align with the rest of the DSL? Suggest we rename this (and the similarly named ones).

kevinrvaz: kept it as matchingRules as i had implemented the same in interaction based methods will update this to withMatchingRules.

kevinrvaz: renamed

src/v4/message/types.ts
mefellows, line 85: I believe we need to add a `withMatchingRules` to the response portion of the Synchronous message also? Asynchronous only has a one way message.

kevinrvaz: added to response as well

kevinrvaz: updated the test to show response matcher as well

PR #1753, opened 2026-04-28
chore: allow script for pact-core

JP-Ellis:
Node package managers like `pnpm` and `aube` disable install scripts by default as a security measure. The `pact-core` dependency should be excepted in our case.


mefellows:
Thanks!

PR #1754, opened 2026-04-28
fix: extract content type from matchers

JP-Ellis:
Fixes failures when consumers use regex matchers for JSON Content-Type headers with optional charset parameters.

Adds regression coverage for the v3 bug and confirms the same scenario already works in v4.

Ref: PACT-6546

### DIFF
--- a/src/v3/ffi.ts
+++ b/src/v3/ffi.ts
@@ -1,6 +1,6 @@
 import { forEachObjIndexed } from 'ramda';
-import { ConsumerInteraction } from '@pact-foundation/pact-core';
-import { Matcher, TemplateHeaders, V3Request, V3Response } from './types';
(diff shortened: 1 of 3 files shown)


JP-Ellis:
Note before merging: this is a patch to the issue. I will take a further look in the FFI

src/v3/ffi.ts
```diff
@@ -59,15 +59,17 @@ export const setResponseDetails = (
   }, res.headers);
 };
 
-// TODO: this might need to consider an array of values
```
mefellows, line 62: 👀

mefellows (approved):
Looks ok to me, I think we can ignore the copilot suggestions unless we are getting lint failures.

PR #1760, opened 2026-04-29
chore: replace eslint+prettier with biome

JP-Ellis:
Replace eslint and prettier with biome as one tool for formatting and linting.


mefellows:
Was waiting for this! Thanks 🙏  - will take a look tonight.

JP-Ellis:
Still have a couple of follow up fixes in the workflow and the docs, and also cleaning up the various examples.

I have split the PR into logical commits, so that will help a lot with the review.

package.json
```diff
@@ -10,14 +10,13 @@
...
-    "lint:fix": "npm run lint -- --fix",
-    "format:base": "prettier \"{src,test,examples}/**/*.{ts,js}\"",
-    "format:fix": "npm run format:base -- --write",
-    "format:check": "npm run format:base -- --check",
+    "lint": "biome lint",
+    "lint:fix": "biome lint --write",
+    "format": "biome format",
+    "format:fix": "biome format --write",
+    "check": "npm run format && npm run lint",
+    "check:fix": "npm run format:fix && npm run lint:fix",
```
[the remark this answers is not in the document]
mefellows: @JP-Ellis can you please double check the scripts match what we do in GH actions?

biome.jsonc
```diff
@@ -0,0 +1,17 @@
+{
```
mefellows, line 1: The vast amount of changes appears to be the single vs double quotes change. 

There are arguments either way, but I suggest we set the default to single quotes - this will reduce the formatting surface and help with `blame` debugging in the future. I don't really care either way, as my assumption is that the linter/formatter will do this for me.

mefellows (changes requested):
Minor changes.

Also suggest updating https://github.com/pact-foundation/pact-js/blob/master/CONTRIBUTING.md which references formatting approach.

src/httpPact/index.ts
```diff
@@ -202,24 +208,24 @@ export class Pact {
 
     // Feature flag: allow missing requests on the mock service
     if (!success) {
-      let error = 'Test failed for the following reasons:';
-      error += `\n\n  ${generateMockServerError(matchingResults, '\t')}`;
+      let error = "Test failed for the following reasons:";
+      error += `\n\n  ${generateMockServerError(matchingResults, "\t")}`;
 
       /* eslint-disable no-console */
```
mefellows: I assume biome knows what this means? Should we update these also?

biome.jsonc
JP-Ellis: I'll set the config to change that over, and rewrite PR history here to fix up the blame (alternatively, I can also create a blameignore file)

package.json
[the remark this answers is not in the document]
JP-Ellis: Other than one `npm run format:check` to `npm run format`, it should be all good.

src/httpPact/index.ts
JP-Ellis: I'm not sure what you mean? This change here was purely a `s/'/"/`. Now that I've configured biome to use single quote, there is no remaining diff.

JP-Ellis:
@mefellows This should be ready for review. I have switched the default quotes to single quotes to minimise the diff.

Not sure why these examples are failing though 🤔

mefellows (approved):

mefellows:
I feel cleaner already - thanks Josh!

PR #1761, opened 2026-04-30
chore: update tsconfig in all examples

JP-Ellis:
Have all examples leverage the root tsconfig, to avoid mismatched configurations. Also added `lint:tsc` to check the types.

Note that mocha doesn't play well with TypeScript (may need to switch to vitest?), and some examples do not have typescript wired in yet. As a result, typescript checks are not enforced in CI, and some examples have `lint:tsc` as a no-op echo.


mefellows:
Seems sensible, it needed a bit of a cleanup.

Let's fix the broken examples (they look to be failing with a TS error) and then we can separately do a cleanup to migrate all examples to TS if we like.

There are historical reasons for the polyglot examples, but I don't think they really add value now. We could probably use the same test runner across all examples also to keep things a little more consistent.

If we want to go one step further, we could rewrite the examples to be more pedagogical to explain all of the key features, rather than the current approach of... I don't know what you'd call it :P

JP-Ellis:
Yeah, one step at a time 😆 

My general idea:

- Let's convert everything to TypeScript. The type annotations help with learning, and if someone wants to adapt an example to TS only, it shouldn't be a huge problem (especially with AI tools now). If there's a genuine need for a JS-only example, then we can do that and make it clear why.
- I want to convert all tests to use `vitest` as the test runner by default. I will keep an example or two featuring a different test runner if there's a real difference.

And yeah, eventually focus on improving the docs, and maybe even cleaning up some old examples (e.g., should we keep the v2 examples? Or should we incentivise moving to v4?)

mefellows:
> I want to convert all tests to use vitest as the test runner by default. 

👍 

>I will keep an example or two featuring a different test runner if there's a real difference.

There isn't. I'm not aware of current problems but they all roughly have the same basic capabilities - anything test framework specific is usually a setup artifact unrelated to Pact itself so I don't think is super useful.

> Let's convert everything to TypeScript. The type annotations help with learning, and if someone wants to adapt an example to TS only, it shouldn't be a huge problem (especially with AI tools now). If there's a genuine need for a JS-only example, then we can do that and make it clear why.

Sounds good. Somebody might complain, but I don't think there is a strong case against doing that.

PR #1768, opened 2026-05-05
feat: add xml support to pactv4

JP-Ellis:
Add support for XML bodies in PactV4. The change should be straightforward as it can leverage the existing xmlBody from PactV3.


JP-Ellis:
This is mostly untested _at the moment_ as I implemented this while reworking the v4 examples. I'll tweak this PR first to add some unit tests outside of the examples.

mefellows:
I've never tested with raw XML strings before. It's possibly it will work.

The V3 interface has an [`XmlBuilder`](https://github.com/pact-foundation/pact-js/blob/master/src/v3/xml/xmlBuilder.ts#L6) which you can use to build up the XML document programmatically. It might be nice to support either a `unknown` or factory function that initiates the builder

Example usage of XML in V3 here: https://github.com/pact-foundation/pact-js/blob/master/examples/v3/e2e/test/consumer.spec.js

JP-Ellis:
Yeah, it came about because I tried to adapt the XML example to Pact V4, but it didn't work 😅

mefellows:
Nice one - thanks!

(now to add support to messages ;) )

PR #1785, opened 2026-05-05
docs: rework examples

JP-Ellis:
Replace the per-version examples with a single suite of examples, all using `Pact` (i.e., `PactV4`).

The old v2 and v3 examples have been converted into the regression suite in #1769.


mefellows:
This is great! I'll have to trust that most examples haven't changed dramatically - there is a lot of churn here, but I'm guessing it's just moving things around and additional commenting (e.g. README and inline clarifications).

There is one failing test here: https://github.com/pact-foundation/pact-js/actions/runs/25367451451/job/74381898185?pr=1785#step:6:743. It looks to be related to broker integration (I haven't looked carefully, but it's possible it's because this is a branch and the version/selectors may not be aligning).

This raises a question - should we keep the broker integration, or keep them local only? 

Either way, the build should fail if a test fails so we should sort that first.

JP-Ellis:
Ah, good catch, I wrongly assumed `set -e` was used for inline bash scripts. I'll fix that. EDIT: it seems like it is used... not sure why it was not causing a quick exit.

To answer your question about the broker integration: I opted _not_ to do that at this stage because it adds a lot of overhead which, for the purposes of an example, don't really contribute much value. So the tests are stored locally only.

Full E2E tests which include the broker I think can best be demonstrated in other places (like our various example repos).

mefellows:
Awesome, thanks!

PR #1789, opened 2026-05-06
chore: migrate to vitest

JP-Ellis:
Replace mocha with vitest for faster test running, and native TS support. Also reduce the number of dependencies by switching to vitest's own assertion library.


JP-Ellis:
PS: Aware of the CI failure, will follow up (just getting it out first for visibility)

PR #1813, opened 2026-05-25
chore!: drop node 20, test against node 26

JP-Ellis:
With Node 20 being now officially unmaintained, drop it from the test matrix and other related fields.


mefellows:
Thanks!

PR #1865, opened 2026-06-25
fix(verifier): run beforeEach/afterEach once per interaction without desyncing on hook failure

### LINKED ISSUE #1864: A failing beforeEach permanently desyncs the hook counter and silently disables beforeEach/afterEach for all later interactions
# A failing `beforeEach` permanently desyncs the hook counter and silently disables `beforeEach`/`afterEach` for all later interactions

### Software versions

- **Provider Pact library**: `@pact-foundation/pact` 16.0.4
- **pact-core**: `@pact-foundation/pact-core` 17.1.0 (native FFI 0.4.28, `pact_verifier` 1.3.x)
- **Node**: 24.x
- **OS**: macOS

### Issue Checklist

- [x] I have read the FAQs in the Readme
- [x] I have triple checked that there are no unhandled promises in my code
- [x] I have set my log level to debug and attached a log showing the request/response cycle
- [x] I have created a reproducible repository

### Expected behaviour

`beforeEach` and `afterEach` should run **once per interaction**, independently for
each interaction. If a hook fails for one interaction, that must not affect whether
the hook runs for subsequent interactions.

### Actual behaviour

A single failing `beforeEach` (e.g. a transient error on the first interaction)
**permanently disables `beforeEach` and `afterEach` for every later interaction** —
and, because of a retry, the verification still reports **success**.

With a 3-interaction pact and a `beforeEach` that throws only on interaction 1:

| | `beforeEach` invoked | `afterEach` invoked | interactions verified | result |
|---|---|---|---|---|
| failing `beforeEach` | **1** (expected 3) | **0** (expected 3) | 3/3 | ✅ passes |
| no failure (control) | 3 | 3 | 3/3 | ✅ passes |

This is silent and dangerous: provider-state/`beforeEach` setup never runs for most
interactions, yet the build is green.

### Steps to reproduce

Minimal repo: https://github.com/frudisch/pact-beforeeach-desync-repro

```bash
npm install
npm run verify            # bug: one transient beforeEach failure
NO_THROW=1 npm run verify # control: no failure
```

Or:

1. A pact with 3 interactions.
2. A `Verifier` whose `beforeEach` throws on its first invocation only.
3. Run `verifyProvider()` with `logLevel: 'debug'`.

### Root cause

The hooks piggyback on the provider-state setup/teardown POSTs to `/_pactSetup`
through a single global counter (`src/dsl/verifier/proxy/hooks.ts`):

```ts
// registerHookStateTracking
if (body?.action === 'setup')    hooksState.setupCounter += 1;
if (body?.action === 'teardown') hooksState.setupCounter -= 1;

// registerBeforeHook -> runs beforeEach when:
body?.action === 'setup'    && hooksState.setupCounter === 1
// registerAfterHook -> runs afterEach when:
body?.action === 'teardown' && hooksState.setupCounter === 0
```

This is correct only if setup/teardown POSTs are perfectly 1:1 balanced for the
whole run. They are not, because:

1. A throwing `beforeEach` calls `next(new Error(...))` → the `/_pactSetup` response
   becomes HTTP **500** (`hooks.ts`).
2. The native core **retries** the failed state-change POST. The retry re-enters
   `registerHookStateTracking` and increments the counter again (`1 → 2`), but
   `beforeEach` is not re-run (it now requires `=== 1`). The retry returns 200 and
   succeeds.
3. The counter floor is now permanently **1**; it oscillates **1↔2** instead of
   **0↔1** for the rest of the run.
4. Every later interaction: `setup` → `1→2` (`beforeEach` requires `1` → skipped);
   `teardown` → `2→1` (`afterEach` requires `0` → skipped).

Debug log excerpt:

```
hooks state counter is 1 after receiving "setup" action      <- I1 setup
executing 'beforeEach' hook
error executing 'beforeEach' hook: transient failure ...      <- 500
hooks state counter is 2 after receiving "setup" action      <- core RETRIES the setup -> desync
hooks state counter is 1 after receiving "teardown" action    <- I1 teardown (floor is now 1)
hooks state counter is 2 after receiving "setup" action      <- I2 setup (beforeEach NOT run)
hooks state counter is 1 after receiving "teardown" action    <- I2 teardown (afterEach NOT run)
hooks state counter is 2 after receiving "setup" action      <- I3 setup (beforeEach NOT run)
hooks state counter is 1 after receiving "teardown" action    <- I3 teardown (afterEach NOT run)
```

More generally, the counter is a global accumulator with no error correction:
**any** unbalanced setup/teardown (a retried setup, a setup whose teardown the core
skips on failure) offsets it permanently.

### Suggested fix

Two complementary changes:

1. **Don't let a hook failure turn into a non-2xx state-change response.** Record
   the hook error and fail `verifyProvider()` at the end instead of `next(err)`.
   This removes the retry/desync trigger entirely.
2. **Replace the global accumulating counter with self-healing, edge-based interaction
   tracking** (e.g. a boolean that flips on the `setup`/`teardown` edges, idempotent
   under retries), so the hooks cannot latch into a permanently-broken state.

### Note — not the same as #924

#924 is the older "stateless interactions are skipped" bug, fixed in the core
(empty-state setup now fires). This is a different defect introduced with the
counter mechanism in #1243: balanced runs work, but any setup retry/failure
permanently desyncs the counter.

frudisch:
## Summary

Fixes the provider-verification `beforeEach`/`afterEach` hooks so they run **once per interaction** and can no longer be silently disabled by a single hook failure.

Closes #1864.

## Problem

The hooks piggyback on the provider-state setup/teardown requests the core sends to `/_pactSetup`, tracked with a single global counter that fires the hooks only on exact equality (`setupCounter === 1` / `=== 0`). This assumes setup/teardown requests stay perfectly balanced for the whole run — they don't:

1. A throwing hook returned a non-2xx `/_pactSetup` response via `next(err)`.
2. The native core **retries** the failed state-change request.
3. The retry re-incremented the counter **without re-running the hook**, permanently offsetting it — the counter then oscillated **1↔2 instead of 0↔1**.
4. From that point `beforeEach` (needs `=== 1`) and `afterEach` (needs `=== 0`) **never fired again**, yet the run still reported **success**.

So one transient hook failure silently skips provider-state setup for every later interaction while the build stays green.

## Reproduction

Minimal repo (pact 16.0.4 / pact-core 17.1.0, FFI 0.4.28): https://github.com/frudisch/pact-beforeeach-desync-repro

3-interaction pact, `beforeEach` throws once on interaction 1:

| | `beforeEach` | `afterEach` | interactions verified | result |
|---|---|---|---|---|
| **before** | 1 (expected 3) | 0 (expected 3) | 3/3 | ✅ passes (silently wrong) |
| **after**  | 3 | 3 | 3/3 | ❌ fails with a clear hook error |
| control (no throw) | 3 | 3 | 3/3 | ✅ passes |

## Fix

- **Edge-based interaction tracking.** Replace the accumulating `setupCounter` with an `insideInteraction` flag that flips on the setup↔teardown boundaries. It is idempotent under retries (a duplicate `setup` while already inside an interaction is a no-op) and cannot latch into a broken state.
- **Decouple hook failures from the state-change response.** A hook that throws no longer produces a non-2xx response (which the core retries). Errors are recorded on the hooks state and surfaced by the verifier once verification completes — so a failing hook still fails the build, with a clear aggregated message, but never desyncs tracking.

## Notes

- `afterEach` now fires on the first teardown of an interaction rather than the last. This is not a behavioural regression: even today `afterEach` runs before the final teardown's state handler completes (middleware ordering), and the interaction's request has already finished before any teardown.
- Relies on the core sending teardown requests (pact-js-core sets `teardown=true`), which is already the case.

## Tests

- Rewrote `hooks.spec.ts` for the new API, adding regression coverage: a throwing `beforeEach` no longer disables hooks for later interactions; a retried `setup` doesn't re-fire or desync; multi-state interactions fire once.
- Full suite green (283 tests).
- Validated end-to-end against the real core (pact 16.0.4 / pact-core 17.1.0) using the repro above.

### DIFF
--- a/src/dsl/verifier/proxy/hooks.spec.ts
+++ b/src/dsl/verifier/proxy/hooks.spec.ts
@@ -1,127 +1,118 @@
 import type { RequestHandler } from 'express';
 import { vi } from 'vitest';
 
-import {
-  type HooksState,
-  registerAfterHook,
-  registerBeforeHook,
-  registerHookStateTracking,
-} from './hooks';
+import { createHooksState, registerHooks } from './hooks';
 
-// This mimics the proxy setup (src/dsl/verifier/proxy/proxy.ts), whereby the
-// state handling middleware is run regardless of whether a hook is registered
-// or not.
-const doRequest = async (
-  action: string,
-  hooksState: HooksState,
-  hookHandler?: RequestHandler,
-) => {
-  const hooksStateHandler = registerHookStateTracking(hooksState);
-  const hookRequestHandler = hookHandler || ((_req, _res, next) => next());
+// Mimics the proxy: a single hooks middleware handles every `/_pactSetup`
+// request. `doRequest` drives it with a given action and resolves once the
+// middleware calls next().
+const doRequest = (action: string, handler: RequestHandler) => {
+  // biome-ignore lint/suspicious/noExplicitAny: partial mock — only body is needed
+  const request: any = { body: { action } };
 
-  // biome-ignore lint/suspicious/noExplicitAny: partial mock object — only body is needed to exercise the hook logic
-  const request: any = {
-    body: {
-      action,
-    },
-  };
-
-  return new Promise((resolve) => {
-    // biome-ignore lint/suspicious/noExplicitAny: null passed as res/next mocks; only body is exercised
-    hooksStateHandler(request, null as any, () => {
-      // biome-ignore lint/suspicious/noExplicitAny: null passed as res mock; only body is exercised
-      hookRequestHandler(request, null as any, resolve);
-    });
+  return new Promise<void>((resolve) => {
+    // biome-ignore lint/suspicious/noExplicitAny: null res mock; only body is exercised
+    handler(request, null as any, () => resolve());
   });
 };
 
-describe('Verifier', () => {
+describe('Verifier hooks', () => {
   afterEach(() => {
     vi.restoreAllMocks();
   });
 
-  describe('#registerBeforeHook', () => {
-    describe('when the state setup routine is called multiple times before the next teardown', () => {
-      it('it executes the beforeEach hook only once', async () => {
-        const hooksState: HooksState = { setupCounter: 0 };
-        const hook = vi.fn().mockResolvedValue(undefined);
-        const hookHandler = registerBeforeHook(hook, hooksState);
+  it('runs beforeEach and afterEach once per interaction (multiple "given" states)', async () => {
+    const state = createHooksState();
+    const beforeEach = vi.fn().mockResolvedValue(undefined);
+    const afterEach = vi.fn().mockResolvedValue(undefined);
+    const handler = registerHooks({ beforeEach, afterEach }, state);
 
-        await doRequest('setup', hooksState, hookHandler);
-        await doRequest('setup', hooksState, hookHandler);
-        await doRequest('teardown', hooksState);
-        await doRequest('teardown', hooksState);
+    // Interaction with two "given" states: setup x2, teardown x2
+    await doRequest('setup', handler);
+    await doRequest('setup', handler);
+    await doRequest('teardown', handler);
+    await doRequest('teardown', handler);
 
-        expect(hook).toHaveBeenCalledOnce();
-      });
-    });
+    expect(beforeEach).toHaveBeenCalledOnce();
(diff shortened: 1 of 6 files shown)


mefellows:
Thanks for raising this. Let me take a good look at it and if the tests are all good we'll get this merged and released!

PR #1947, opened 2026-09-16
feat(verifier): structured verification results and per-interaction test suites

frudisch:
## Summary

A provider verification is currently a single opaque step. `verifyProvider()` resolves with the raw output of the core and rejects on failure, so a test framework reports **one** test case no matter how many pacts and interactions the run covered. When it fails, the reader has to search the log to find out which interaction broke.

This PR adds two things on top of the existing API, without changing `verifyProvider()`:

1. **`Verifier.verify()`** runs the same verification and resolves with a typed `VerificationResult` holding one entry per verified interaction.
2. **`defineVerificationSuite(options, { describe, it })`** turns the verification into **one `describe` per pact and one `it` per interaction** for Jest, Vitest or Mocha.

```
 ❯ provider.test.ts (5 tests | 3 failed)
   ✓ DemoConsumer -> DemoProvider > interaction 1 - GET /one
   × DemoConsumer -> DemoProvider > interaction 2 - GET /two
     → Verification of 'DemoConsumer -> DemoProvider: interaction 2 - GET /two' failed:
         - BodyMismatch at $: Actual map is missing the following keys: ok
         - StatusMismatch: expected 200 but was 500
   × DemoConsumer -> DemoProvider > interaction 3 - GET /three
   ✓ OtherConsumer -> DemoProvider > a request for one
   × OtherConsumer -> DemoProvider > a request for two
```

Failures show up next to the interaction that caused them, and CI reports (JUnit and friends) list every interaction as its own test case.

## `verify()`

```js
const { Verifier, ProviderVerificationError, groupByPact } = require('@pact-foundation/pact');

try {
  const result = await new Verifier(opts).verify();
  for (const pact of groupByPact(result)) {
    console.log(pact.consumer, '->', pact.provider, pact.interactions.length);
  }
} catch (e) {
  if (e instanceof ProviderVerificationError) {
    for (const i of e.result.interactions.filter((x) => !x.success)) {
      console.log(i.description, i.failure);
    }
  } else {
    throw e; // the verification could not run at all (configuration, hooks, core crash)
  }
}
```

Each `InteractionVerificationResult` has `description`, `success`, `pending`, `duration` and, for failures, `failure` with the mismatches or the error message. `consumer`, `provider` and `providerStates` are reported by the core since `@pact-foundation/pact-core` 20.2.0 ([pact-js-core#957](https://github.com/pact-foundation/pact-js-core/pull/957), [pact-reference#549](https://github.com/pact-foundation/pact-reference/pull/549)). The parser also understands the previous format, where it recovers the pact and the mismatches of failed interactions from the `errors` list, so the feature works with pact-core 19 and 20.1.x too.

## `defineVerificationSuite()`

```ts
import { describe, it } from 'vitest'; // or the Jest/Mocha globals
import { defineVerificationSuite } from '@pact-foundation/pact';

describe('UserProvider', () => {
  let server;
  let baseUrl = '';
  beforeAll(async () => { /* start provider, set baseUrl */ });
  afterAll(() => server.close());

  defineVerificationSuite(
    {
      provider: 'UserProvider',
      get providerBaseUrl() { return baseUrl; },
      pactUrls: [path.resolve(process.cwd(), 'pacts')],
      stateHandlers: { /* ... */ },
    },
    { describe, it },
  );
});
```

Design notes:

- The pact files are read **synchronously while the suite is defined**, because `describe`/`it` cannot be registered from inside a running test. Only local pact files or directories can be enumerated; `pactFiles` lets you keep local copies for reporting while verifying from a broker.
- The verification runs **once**, when the first test case executes. All test cases share that run, so there is no per-interaction re-verification and broker publishing stays correct. `timeout` (default 30 s) has to cover the whole run.
- Verifier options are read when the run starts, so a provider started in `beforeAll` on an ephemeral port can be supplied through a getter.
- Results are matched to pact file interactions by interaction key, then by (consumer, provider, description), then by description and position.
- Pending interactions never fail a test case; a missing result (e.g. filtered out via `PACT_DESCRIPTION`) does.
- The adapter only needs `describe(name, fn)` and `it(name, fn, timeout)`, so pact-js stays framework-neutral and gains no test framework dependency.

## Changes

- `src/dsl/verifier/result/`: result types, the parser (`parseVerificationResult`, `groupByPact`), `ProviderVerificationError`, and a synchronous pact file reader (V2 `providerState`, V3/V4 `providerStates`, message pacts)
- `src/dsl/verifier/suite.ts`: `defineVerificationSuite`
- `src/dsl/verifier/verifier.ts`: `verify()` next to the unchanged `verifyProvider()`
- `src/index.ts`: exports
- `docs/provider.md`: sections "Structured Verification Results" and "Reporting each Pact and Interaction as a Test Case"
- `examples/verification-results/`: two consumers against one provider, one test file using the suite and one keeping the conventional `verifyProvider()` call, plus a script that runs against a deliberately broken provider

## Test plan

- [x] `npm test` — 337 passed (31 files), including 37 new specs for the parser, the pact file reader, the suite adapter and `verify()`
- [x] `npm run lint` (biome + tsc) — clean
- [x] Parser specs run against real verifier output captured from pact-core, in both the current and the previous JSON format
- [x] `examples/verification-results`: consumer tests, the suite and the baseline test, against a healthy and a deliberately failing provider, with pact-core 20.1.1 and 20.2.0

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BdhPcAh5tepommAaXELEFw


frudisch:
Fixed the CI failures in the first run, both were real problems in this branch:

1. **Test fixtures were never committed.** The `pacts/` entry in `.gitignore` matches any directory with that name, so `src/dsl/verifier/__fixtures__/pacts/` was silently excluded and the new specs had no input in CI. The directory is now `__fixtures__/pact-files/` and its files are tracked. `suite.spec.ts` also read the pact files generated by `examples/verification-results` (likewise gitignored); it now uses committed copies, so the specs no longer depend on generated output.
2. **Format check.** Two JSON fixtures were not biome-formatted.

Verified in a clean checkout of the pushed commit (`git worktree` from the commit, `npm ci`): `npm run format`, `npm run lint` and `npm test` (337 passed) are green, and `examples/verification-results` passes `npm ci && npm test`.

frudisch:
Extended the tests, and fixed two things they uncovered:

- **A suite that would register nothing now throws.** If the sources contain no pact file, or the pact files contain no interactions, `defineVerificationSuite` raises a `ConfigurationError` instead of registering zero test cases. Silently registering nothing made a run that verified nothing look green, which is exactly the failure mode contract tests exist to prevent.
- **Filesystem and JSON errors name their source.** A missing path used to surface as a raw `ENOENT` and a malformed pact file as a bare `SyntaxError` with no filename, which is unhelpful when a whole directory is read. Both are now `ConfigurationError`s naming the source.

New test coverage: the two guards above, broker-only attributes (`interactionId`, `interactionKey`), header and query mismatch shapes, broker notices, message pacts in a suite, provider states given as plain strings, malformed result documents, failure descriptions that do not follow the usual pattern, the positional fallback when a reported pact no longer matches the pact file, and `ProviderVerificationError` message construction.

| file | branch coverage before | after |
| --- | --- | --- |
| `result/parse.ts` | 78.1% | 88.9% |
| `result/pactFiles.ts` | 89.2% | 93.2% |
| `result/error.ts` | 85.7% | 100% |
| `suite.ts` | 94.8% | 98.5% |

Verified again in a clean checkout of the pushed commit: `npm run format`, `npm run lint`, `npm test` (357 passed, 32 files) and `examples/verification-results` with `npm ci && npm test`.

mefellows:
This looks great, thanks for starting this. 

I'll provide more detailed review (likely next week) but at first pass I'm not sure I'm comfortable with `defineVerificationSuite` only supporting local pact files. The most common use case is use with a broker, so if we can't support that here I'm not really sure of its value. Downloading a pact file in advanced for this use case breaks the linkages in the network responses to ensure verification results are properly linked, so it really prevents the majority of heavy users from getting the value. 

But returning structured results is a fantastic addition and will make output a lot nicer.

frudisch:
You are right, and thank you for naming it early. Verifying a downloaded pact file does lose the broker linkage: `publish_result` in `pact_verifier` only publishes for the `BrokerUrl` and `BrokerWithDynamicConfiguration` sources and logs "not possible" for everything else, and `interaction_id` is only set for pacts loaded from a broker. A suite that only works on local files would indeed miss the users who need this most.

I have pushed broker support. The fix is to stop enumerating pact files and take the test cases from the verification result instead — which is exactly what the pact-core 20.2.0 change enables, since every interaction result now names its consumer and provider:

```ts
import { describe, it } from 'vitest';
import { defineVerificationSuiteAsync } from '@pact-foundation/pact';

const server = createApp().listen(3001);

await defineVerificationSuiteAsync(
  {
    provider: 'UserProvider',
    providerBaseUrl: 'http://localhost:3001',
    pactBrokerUrl: process.env.PACT_BROKER_BASE_URL,
    consumerVersionSelectors: [{ mainBranch: true }, { deployedOrReleased: true }],
    publishVerificationResult: true,
    providerVersion: process.env.GIT_COMMIT,
  },
  { describe, it },
);
```

This runs an ordinary broker verification, so the links stay intact and publishing keeps working. Every pact the broker returned becomes a `describe` and every interaction an `it`, pending and WIP pacts included, with no second round trip and no local copies that could drift from the selectors.

The one trade-off is timing: a test framework has to know its test cases before it starts running them, so the verification has to finish first. That means the provider must be reachable when the call is awaited, and the provider is started at the top level rather than in `beforeAll`. Vitest and Mocha support awaiting at the top level; for Jest there is `verifyForSuite()` to run in a `globalSetup` and `defineVerificationSuiteFromResult()` to register the cases synchronously in the test file. All three are documented in `docs/provider.md`.

Other changes in this push:

- `@pact-foundation/pact-core` bumped to `^20.2.0`. Grouping by pact needs the consumer and provider per interaction; with older cores the interactions whose pact is unknown land in one shared group and a warning is logged.
- `examples/verification-results/provider-async.test.ts` demonstrates the result-driven variant end to end.
- 11 new specs for the result-driven API, including that the broker options reach the verifier and the naming callbacks do not.

`defineVerificationSuite()` (the pact-file variant) is still there for the local-file case, but I am happy to drop it if you would rather have one way to do this. Clean checkout of the pushed commit: format, lint and `npm test` (368 passed) green, example green.

frudisch:
Pushed a fix for the Windows failures. They were mine: a new spec built a regular expression from a temp file path to assert that the error message names the file, and on Windows the backslashes in `C:\Users\...` are escape sequences. Both path assertions now match the path as a substring instead.

The remaining red job, `Test on ubuntu-latest with Node 22`, is unrelated to this branch: `src/pact.integration.spec.ts > Using the MATT plugin > Synchronous Message (TCP) > generates a pact` timed out after 60s. The same file is untouched here, the same test passes on the other Node versions in that run, and the Test workflow is green on master. Looks like a flake in the plugin/TCP setup; the new run should tell.

